PR #51, opened 2015-11-13
Unneccessary async/await, typos

schjan:
#### async/await

You get no benefit of awaiting here, because this function doesn't do anything async, so you could easily return the Task (awaiter) of the function instead of awaiting it.

```csharp
// old awaits SendMessage(...)
public async Task<Message> SendDocument(string chatId, string document, int replyToMessageId = 0,
      ReplyMarkup replyMarkup = null) =>
            await SendMessage(MessageType.DocumentMessage, chatId, document, replyToMessageId,
                  replyMarkup).ConfigureAwait(false);

// new: returns awaitable Task of SendMessage(...)
public Task<Message> SendDocument(string chatId, string document, int replyToMessageId = 0,
      ReplyMarkup replyMarkup = null) =>
            SendMessage(MessageType.DocumentMessage, chatId, document, replyToMessageId,
                  replyMarkup);

```

### typos

There was a typo in the optional parameter __isMarkdown__ of the _SendTextMessage_ method. This results in a possible breaking change.

There was a typo in Message.__ReplyToMessage__, which is also a breaking change.

### DIFF
--- a/Telegram.Bot/Api.cs
+++ b/Telegram.Bot/Api.cs
@@ -26,7 +26,7 @@ namespace Telegram.Bot
         /// A simple method for testing your bot's auth token. Requires no parameters. Returns basic information about the bot in form of User object.
         /// </summary>
         /// <returns>Returns basic information about the bot in form of <see cref="User"/> object</returns>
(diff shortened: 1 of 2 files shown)


MrRoundRobin:
thx!

PR #91, opened 2016-02-15
removing parameter.Key == "timeout" because timeout in the HttpClient

juanl:
Hi there. 
There is a problem with the timeout parameter used in the getUpdates method. When you send 0 (default value), the HttpClient timeout is set to 1 secs.


MrRoundRobin:
should be fixed in f8ebeaf85e76dec965e1f6e74687341af8a292ba

PR #105, opened 2016-03-10
fixed UnixDateTimeConverter

dyarovyi:
For .NET 4.6.1 there were a problem with Message.Date property: after deserializing it contained only Date part, but not Time (so all messages, received in the same day, had the same value of this property)


MrRoundRobin:
sorry for the delay, there is an even simpler fix, see 78f45f6

PR #117, opened 2016-04-12
Added some inline keyboard and edit message features

astarcev:


inornos:
how to use inline keyboard please share the code

PR #118, opened 2016-04-14
Resolved some issues with the InlineResultTypes

18padx08:
Just removed the duplicated entry and added the one i needed


dedo1911:
@MrRoundRobin I confirm this also should be merged

PR #123, opened 2016-04-22
API 2.0: Added optional location to Inline query

dedo1911:


dedo1911:
Need more testing.

dedo1911:
Works as expected :+1:

Issue #131, opened 2016-04-29
InlineKeyboardMarkup isn't a ReplyMarkup


adebisi-fa:
**InlineKeyboardMarkup** can't be used with methods that accept **ReplyMarkup**, e.g. api.**SendTextMessage**(...), etc.

PR #132, opened 2016-04-29
Fixes issue #131

adebisi-fa:
- **ReplyMarkup** and **InlineKeyboardMarkup** implement the **IReplyMarkup** interface.
- API methods that previously support **ReplyMarkup**, now require **IReplyMarkup** interface instead.

These, thus, allow **InlineKeyboardMarkup** be used with all API methods that previously accept **ReplyMarkup**.


MrRoundRobin:
i worked on silent messages since yesterday, so now there are conflicts.

but the code look's good, i'll do it quickly with s&r, faster then resolving the conflicts

adebisi-fa:
That's OK.

Thanks for the great effort you're putting into getting this library up to date. 
It is the best around, in the .Net space!

Well, done!

PR #196, opened 2016-07-05
Fixes #195

### LINKED ISSUE #195: Unable to edit message without inline keyboard.
### Steps to reproduce
Send a message without an inline keyboard with a text. Immediately after that, update the message to another text, again without an inline keyboard.

### Expected behavior
The message gets changed to the new text.

### Actual behavior
`Bad request: object expected as reply markup`

### Steps to fix
Use the same check for `replyMarkup` as is used for `parseMethod`.

mhverbakel:
Fixes #195


MrRoundRobin:
nice, i'll add the check to `EditInlineMessageTextAsync`, `EditMessageCaptionAsync`, `EditInlineMessageCaptionAsync`, `EditMessageReplyMarkupAsync` and `EditInlineMessageReplyMarkupAsync`

PR #215, opened 2016-08-11
Patch DateTime min value for JsonConvert

parabola949:
If server time zone has a positive UTC timezone (UTC +3 for example), using `JsonConvert.SerializeObject()` on a CallbackQuery will throw am `ArgumentOutOfRangeException` with the message 

> The UTC time represented when the offset is applied must be between year 0 and 10,000.

1970 is the DateTimeOffset.MinValue, which doesn't cause any errors with this.

### DIFF
--- a/src/Telegram.Bot/Converters/UnixDatetimeConverter.cs
+++ b/src/Telegram.Bot/Converters/UnixDatetimeConverter.cs
@@ -29,9 +29,9 @@ namespace Telegram.Bot.Converters
             if (value is DateTime || value is Nullable<DateTime>)
             {
 #if NET46
-                val = new DateTimeOffset((DateTime)value).ToUnixTimeSeconds();
+                val = (DateTime) value == DateTime.MinValue ? -62135596800L : new DateTimeOffset((DateTime)value).ToUnixTimeSeconds();
 #else
-                val = ((DateTime)value).ToUnixTime();
+                val = (DateTime) value == DateTime.MinValue ? -62135596800L : ((DateTime)value).ToUnixTime();
 #endif
             }
             else


src/Telegram.Bot/Converters/UnixDatetimeConverter.cs
```diff
@@ -29,9 +29,9 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
             if (value is DateTime || value is Nullable<DateTime>)
             {
 #if NET46
-                val = new DateTimeOffset((DateTime)value).ToUnixTimeSeconds();
```
MrRoundRobin: isn't it enough to change this line to 
`val = new DateTimeOffset(((DateTime)value).ToUniversalTime()).ToUnixTimeSeconds();
`

src/Telegram.Bot/Converters/UnixDatetimeConverter.cs
```diff
@@ -29,9 +29,9 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
+                val = (DateTime) value == DateTime.MinValue ? -62135596800L : new DateTimeOffset((DateTime)value).ToUnixTimeSeconds();
 #else
-                val = ((DateTime)value).ToUnixTime();
```
MrRoundRobin: and this line to
`val = ((DateTime)value).ToUniversalTime().ToUnixTime();`
?

MrRoundRobin:
@parabola949 can you reply to my comments?

MrRoundRobin:
should be resolved with the latest commits

PR #217, opened 2016-08-14
OnReceiveGeneralError for all exceptions

CristianT:
The ReceiveAsync function is making the host application crash when there is a network error, for example a disconnection.

I propose catching the exception and raising an event to let the user decide what to do with it.

The recommendation would be to call StopReceiving and probably call StartReceiving at a later time to do a reconnection.

### DIFF
--- a/src/Telegram.Bot/Args/ReceiveGeneralErrorEventArgs.cs
+++ b/src/Telegram.Bot/Args/ReceiveGeneralErrorEventArgs.cs
@@ -15,10 +15,10 @@ namespace Telegram.Bot.Args
         /// <value>
         /// The exception.
         /// </value>
-        public Exception Exception { get; private set; }
+        public Exception Exception { get; }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="ReceiveErrorEventArgs"/> class.
+        /// Initializes a new instance of the <see cref="ReceiveGeneralErrorEventArgs"/> class.
         /// </summary>
         /// <param name="exception">The general exception.</param>
         internal ReceiveGeneralErrorEventArgs(Exception exception)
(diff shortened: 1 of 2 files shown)


mhverbakel:
In what context would a simple reconnect not be the solution?

src/Telegram.Bot/Args/ReceiveGeneralErrorEventArgs.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <seealso cref="EventArgs" />
+    public class ReceiveGeneralErrorEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Gets the exception.
+        /// </summary>
+        /// <value>
+        /// The exception.
+        /// </value>
+        public Exception Exception { get; private set; }
```
mhverbakel: Private set can be removed as it is only set from within the constructor

src/Telegram.Bot/Args/ReceiveGeneralErrorEventArgs.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ReceiveErrorEventArgs"/> class.
```
mhverbakel: ReceiveGeneralErrorEventArgs

CristianT:
The exception being catched is too generic to simply reconnect when receiving the event. In some cases a reconnection would suffice, but in some other cases you could enter in an infinite loop until the problem is resolved.

For example in my tests I closed the connection with the telegram server using tcpview (I'm on windows).
As that is an artificial break, a simple reconnect happily continues with the communication.
If you disconnect the network cable however, a call to StartReceiving would intermediately raise the event again so you need to do some sort of retry logic to avoid CPU saturation.

MrRoundRobin:
hi,

my plan was to implement this a little bit different including automatic retries. but i'm currently very bussy so this has too wait an I'll merge this PR.

thanks for your work!

PR #244, opened 2016-10-12
Running under .NET Core

pushist1y:
I've tried to compile and run this code under .NET Core and it worked with no changes to codebase.


MrRoundRobin:
Hi @pushist1y,

that's not needed as netstandard 1.1 is supported by netcoreapps.
so it's already possible to use this library in core apps

PR #257, opened 2016-11-16
Assign url value to InlineKeyboardButton in constructor.

a-a-k:
I saw that it was implemented earlier but I can't find a reason why it was removed. Could you merge it please?


mhverbakel:
This change will not be backwards compatible **without recompiling the code**. To ensure backwards compatibility, add a new overload with a non-optional parameter.

a-a-k:
@mhverbakel check it please.

mhverbakel:
Are you sure this works? I believe that ctor(string, string) is now ambiguous... The last parameter shouldn't be optional if I'm not mistaking.

a-a-k:
> The last parameter shouldn't be optional

It's not necessary. In this case `ctor(string, string)` the next ctor `public InlineKeyboardButton(string text, string callbackData = null)` will be executed.    
But you are right that `ctor(string, string)` is ambiguous for understanding. I've changed this. Check it please.

mhverbakel:
You have my blessing!

PR #258, opened 2016-11-16
add switch_inline_query_current_chat to InlineKeyboardButton

spaceorc:
add switch_inline_query_current_chat to InlineKeyboardButton according to https://core.telegram.org/bots/api/#inlinekeyboardbutton


mhverbakel:
Seems good to me!

PR #316, opened 2017-03-03
Override method Equals for User

Quantum-0:


mhverbakel:
I disapprove this. Only checking the id property is not enough, because two user objects with different properties might be deemed unequal in some existing applications, which would be broken by them suddenly be equal.

That aside, overriding equals also means overriding GetHashCode.

Olfi01:
i actually agree with mhverbakel, two user objects aren't the same if they have the same ID. If you wanna check whether it's the same person, just go on and compare the IDs yourself.

Olfi01:
@Quantum-0 how about this?
```c#
User.Is(User u) => u.Id == this.Id
```

pouladpld:
How about adding a few useful extension methods for frequent comparisons instead?

```c#
// extension methods in Telegram.Bot namespace

bool isSameUser = someUser.IdEquals(otherUser);
bool hasSameSender = someUpdate.IsFromSameUserId(otherUser); // for messages, channel posts, callback query,...
int? msgId = someUpdate.TryGetMessageId(); // for messages, channel posts, callback query,...
long? chatId = someUpdate.TryGetChatId(); // for messages, channel posts, callback query,...
string idOrNull = someUpdate.TryGetCallbackQueryId();
```

Olfi01:
I guess this can be closed? I have opened #442 for discussion instead

pouladpld:
@Quantum-0 I'm closing this PR as we planning to add extension methods for this purpose. Feel free to join our discussion in #442

PR #360, opened 2017-06-23
Fixes #325 (fixed inline message editing)

### LINKED ISSUE #325: Exception in EditInlineMessageTextAsync
Hi 
i use EditInlineMessageTextAsync to edit the inline message from user. 
this function works correctly but at the end of execute send exception that say
"Error converting value True to type 'Telegram.Bot.Types.Message'."

i think exeption throw when try to convert Boolean response to Task<Message> in EditInlineMessageTextAsync method.

Thanks

rgbweb:
Fixes JSON parse exceptions while parsing the bool result from Telegram in the following methods: EditInlineMessageTextAsync, EditInlineMessageCaptionAsync and EditInlineMessageReplyMarkupAsync

### DIFF
--- a/src/Telegram.Bot/ITelegramBotClient.cs
+++ b/src/Telegram.Bot/ITelegramBotClient.cs
@@ -23,7 +23,7 @@ namespace Telegram.Bot
         /// Timeout for requests
         /// </summary>
         TimeSpan Timeout { get; set; }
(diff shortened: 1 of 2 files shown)


MrRoundRobin:
included this in the commits for api 3.1

thx

PR #362, opened 2017-06-26
User.Id and Chat.Id reverted to long

buldo:
According to Telegram [documentation](https://core.telegram.org/bots/api#chat) Id of Chat and User can be only Integer.

### DIFF
--- a/src/Telegram.Bot/Types/Chat.cs
+++ b/src/Telegram.Bot/Types/Chat.cs
@@ -13,7 +13,7 @@ namespace Telegram.Bot.Types
         /// Unique identifier for this chat, not exceeding 1e13 by absolute value
         /// </summary>
         [JsonProperty(PropertyName = "id", Required = Required.Always)]
-        public ChatId Id { get; set; }
+        public long Id { get; set; }
 
         /// <summary>
         /// Type of chat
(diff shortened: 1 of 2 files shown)


pouladpld:
So we do not need `ChatId` type any `long`er :smile:. 
Go ahead and remove ChatId as this is going to be a breaking change.

buldo:
@pouladpld we need ChatId type for methods like [SendMessage](https://core.telegram.org/bots/api#sendmessage).
This method takes Integer or String as the chat_id

Ramin-Bateni:
I wrote more detail [here](https://github.com/MrRoundRobin/telegram.bot/issues/361#issuecomment-312441713) about this issue.

Olfi01:
maybe think about adding an implicit conversion of long or string into ChatId?

Olfi01:
close this PR, as the ChatId type is useful and @dllz has got a PR for the implicit operators, as this will solve your problems.

Olfi01:
@buldo

Olfi01:
UM, ACTUALLY nevermind. xD
If using redis, storing ChatIds is not really possible by itself but if you combine this and #351 you could store chat and user objects as well as still have the useful ChatId type without any problem :smile: doesn't that make everyone happy?
Im still discussing with @dllz tho xD

Olfi01:
well, since @dllz is storing chatIds themselves the problem would still be present. Nevermind.

dllz:
Yeah its fine, I will just have to modify the dll to add an implicit operator for RedisValue to return long. 

Unless anyone has another idea to make it return long by default if it does not have an operator defined for that type?

MrRoundRobin:
So I finally had time to think about this.

I decided to leave chatId for methods and revert the id properties on the chat and user type

i think this is the most comfortable (with the implicit converters) solution

PR #370, opened 2017-07-01
Override ChatId.Equals method

a-a-k:
Comparing by Identifier or Username. Identifier should be nullable.


pouladpld:
@a-a-k I believe this is not necessary anymore since we only use `ChatId` in method parameters. Do you think we can close this PR?

PR #390, opened 2017-07-14
Fixed getChatMemberAsync failing when trying to get a restricted member

Olfi01:


Olfi01:
I basically just added one member to the ChatMemberStatus enum

PR #391, opened 2017-07-15
Made it possible to create InlineKeyboardButtons without callback data

Olfi01:
This way it's now easier again to create InlineKeyboardButtons that switch to an inline query, because otherwise the callback data would just override them.


Olfi01:
I just discovered the .WithSwitchInlineQuery thingies, but in my personal opinion we should leave the user the choice of creating those by himself

MrRoundRobin:
I'd reather create child classes like `InlineKeyboardCallBackButton` and `InlineKeyboardPayButton` and make the `InlineKeyBoardButton` abstract

Olfi01:
That's actually not a bad idea... I'll try :)

MrRoundRobin:
please create a new clean pull request for this

Olfi01:
okay

PR #392, opened 2017-07-15
Added new classes for InlineKeyboardButtons and made InlineKeyboardButton abstract

Olfi01:


MrRoundRobin:
can you move the properties to the child classes

Olfi01:
If you want... but in my opinion the class InlineKeyboardButton should still have all those because you maybe get to Deserialze one and don't wanna check the type, just get the properties, it'd be easier... plus, doing so could maybe meddle with JsonConvert

MrRoundRobin:
that's possible, it's the same as the InlineQueryResults

Olfi01:
okay, I'll take a look

PR #395, opened 2017-07-19
Fix #372

### LINKED ISSUE #372: SetGameScore response is not always a Message
### Steps to reproduce
As a user, not bot, share bot's game to a chat. This means message is not sent by bot itself using SendGame method.
User starts playing the game in that chat and the following code sets user score:

```c#
int userid = 900000000;
string inlineMessageId = "27abCdEFGHIJkLMNOPQRSTuVWX";
int score = 1;
Message response = await botClient.SetGameScoreAsync(userid, score, inlineMessageId);
```

### Expected  behavior
Response should not be a `Message` but rather just a `bool` with value of `true`. The reason is that this game message that user clicked on to play is not sent by bot itself.
In any other situation, the response contains edited `Message`.
More info at <https://core.telegram.org/bots/api#setgamescore>.

### Actual behavior
Throws exception in `TelegramBotClient.cs` at _line 1621_ while trying to deserialize response payload into a `Message`.

```c#
/*
Type of T is Message.
Value of responseString is:
   {"ok":true,"result":true}
*/
responseObject = JsonConvert.DeserializeObject<ApiResponse<T>>(responseString);
```

#### NuGet Package Version
12.0.0-beta-04

#### .NET Version
.NET Core 1.1

#### IDE
VS 2017

pouladpld:


src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -1573,7 +1573,28 @@ public Task<Chat> GetChatAsync(ChatId chatId, CancellationToken cancellationToke
...
+                    message = default(Message);
+                else
+                    throw new ApiRequestException("API responded with false");
+            }
+            else
+            {
+                throw new Exception("Unexpected response");
+            }
+
+            return message;
```
MrRoundRobin: can we Change this to something like this:
```csharp
switch (response)
{
    case Message message:
        return message;
    case bool b:
        if (b)
            return null;
        throw new ApiRequestException("whatever");
    default:
        throw new ApiRequestException("whatelse"); 
}

```

see: https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching#using-pattern-matching-switch-statements

pouladpld: good idea. new fancy features

pouladpld: made the changes

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -1573,7 +1573,22 @@ public Task<Chat> GetChatAsync(ChatId chatId, CancellationToken cancellationToke
...
+                case Message m:
+                    message = m;
+                    break;
+                case bool b when b:
+                    message = default(Message);
+                    break;
+                default:
+                    throw new Exception($"Unexpected response: {response}");
+            }
+            return message;
```
MrRoundRobin: - you can return directly in the case
- a more specific exception would be nice
- use var instad of object

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -1744,7 +1759,7 @@ public Task<bool> DeleteChatPhotoAsync(ChatId chatId, CancellationToken cancella
             };
 
             if (disableNotification)
-                parameters.Add("disable_notification", disableNotification);
+                parameters.Add("disable_notification", true);
```
MrRoundRobin: why?

pouladpld: Just to make resharper happy. Bad habit 😄  Should I revert it?

src/Telegram.Bot/TelegramBotClient.cs
pouladpld: What type of exception is appropriate for case of `false` value and default case?

pouladpld: I don't believe any of those cases will actually happen. I expect tg to respond with error rather than `false` or any other unexpected value.

pouladpld:
@MrRoundRobin did you get a chance to review new changes?

MrRoundRobin:
@pouladpld have you tested the Switch?

pouladpld:
Of course I did

PR #399, opened 2017-07-23
Update to Bot API 3.2

Olfi01:


src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -1767,6 +1767,143 @@ public Task<bool> UnpinChatMessageAsync(ChatId chatId, CancellationToken cancell
...
+            {
+                { "user_id", userId },
+                { "name", name },
+                { "title", title },
+                { "png_sticker", pngSticker },
+                { "emojis", emojis }
+            };
+
+            if (isMasks)
+                parameters.Add("is_masks", isMasks);
```
pouladpld: There is a `contains_masks` property according to [Tg docs](https://core.telegram.org/bots/api#createnewstickerset). Is this a mistake in docs or here?

_Unfortunately, I cannot test it right now._

src/Telegram.Bot/Types/StickerSet.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// Sticker set title
+        /// </summary>
+        [JsonProperty(PropertyName = "title", Required = Required.Always)]
+        public string Title { get; set; }
+
+        /// <summary>
+        /// True, if the sticker set contains masks
+        /// </summary>
+        [JsonProperty(PropertyName = "is_masks", Required = Required.Always)]
+        public bool IsMasks { get; set; }
```
pouladpld: and here

src/Telegram.Bot/TelegramBotClient.cs
Olfi01: here. dang. just got back from a journey.

PR #415, opened 2017-07-31
Fix "Bad Request: file must be non-empty" error when use stream as a source for SendPhotoAsync

suglosta:
When you generates image on the fly (i use OxyPlot to generate charts) and send it using SendPhotoAsync, you will get an error: Bad Request: file must be non-empty.
To avoid this situation you need to set Position property of that stream to 0.
I am proposing to add this to the FileToSend constructor.
It should not affect any other cases e.g. when you reading stream from the file on the disk.

### DIFF
--- a/src/Telegram.Bot/Types/FileToSend.cs
+++ b/src/Telegram.Bot/Types/FileToSend.cs
@@ -54,6 +54,7 @@ namespace Telegram.Bot.Types
         {
             Filename = filename;
             Content = content;
+            Content.Position = 0;	 
 
             Url = null;
             FileId = null;


pouladpld:
@suglosta I haven't used OxyPlot. Could you give me a sample code to test this?

pouladpld:
OK. I now get what causes the problem.

```c#
using (var stream = new MemoryStream())
{
    var writer = new StreamWriter(stream);
    writer.Write("Hello World");
    writer.Flush();
    stream.Position = 0; // causes issue
    await BotClient.SendDocumentAsync(_fixture.SuperGroupChatId, stream.ToFileToSend("file.txt"));
}
```

PR #418, opened 2017-08-04
Fix minor issue with unix time calculation.

Truil:
The main problem was with delta calculation.
Incorrect: var delta = **dateTime** - UnixStart;
Correct: var delta = **utcDateTime** - UnixStart;

Also UnixStart is better to declare with DateTimeKind:
UnixStart = new DateTime(1970, 1, 1, 0, 0, 0, **DateTimeKind.Utc**);


pouladpld:
@Truil Thank you for the PR. Is there any issue there except bad naming?

Truil:
@pouladpld No, I did not notice other issues.

PR #443, opened 2017-08-25
Wikis

pouladpld:
There are some guidelines for contributes. Think of this as a draft and we will keep editing them.


pouladpld:
@Olfi01 No. CONTRIBUTING file should be in the root of repo. That's how gh works ([see this](https://opensource.guide/starting-a-project/#writing-your-contributing-guidelines)).

I don't like to put information on gh wikis because that's technically another git repo and usually goes out of sync with the project. That's why I think it's better to have `/docs/wiki` directory in repos to make finding wiki changes related to each code change easier.

PR #458, opened 2017-09-08
Request enhancement

Olfi01:
This update enables users of webhooks to answer requests with other requests and changes the behaviour of TelegramBotClient to avoid double code, as discussed with @pouladpld

### DIFF
--- a/src/Telegram.Bot/TelegramBotClient.cs
+++ b/src/Telegram.Bot/TelegramBotClient.cs
@@ -21,6 +21,7 @@ using Telegram.Bot.Types.Payments;
 using Telegram.Bot.Types.ReplyMarkups;
 
 using File = Telegram.Bot.Types.File;
+using Telegram.Bot.Types.Requests;
 
 namespace Telegram.Bot
 {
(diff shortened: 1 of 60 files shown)


MrRoundRobin:
Hi, i like the idea! 

I tought about the webrequest problem long time. I always imagined simeting like this:

```csharp

TelegramBot.SendMessage().Invoke(); // To send request
TelegramBot.AnswerInlineQuery().Invoke();
// vs.
TelegramBot.SendMessage().GetStream() // to get response stream
TelegramBot.AnswerInlineQuery();
```

with that we can easily extend with global options like

```csharp
TelegramBot.SendMessage().ReplyMarkup().Invoke()
TelegramBot.SendMessage().ForceReply().Invoke()
``` 

you get the Idea?

i think this is much more easy to extend even outside of the library it self and you dont need that much classes.

sorry for the late reply

pouladpld:
@Olfi01 Ty for this work but using dictionary doesn't offer much reusability. The work with requests of `IRequest<TResult> where TResult : IResponse` has already began and using `RequestBase`  allows us to reuse many similar params across requests.

I don't believe we go with this approach any more.

PR #488, opened 2017-10-16
Implement Bot API v3.4

Tuscen:


src/Telegram.Bot/Types/Chat.cs
```diff
@@ -74,5 +74,17 @@ public class Chat
         /// </summary>
         [JsonProperty(PropertyName = "pinned_message", Required = Required.Default)]
         public Message PinnedMessage { get; set; }
+
+        /// <summary>
+        /// Optional. For supergroups, name of group sticker set. Returned only in getChat.
+        /// </summary>
+        [JsonProperty(PropertyName = "sticker_set_name", Required = Required.Default)]
```
pouladpld: use naming strategy instead.

```c#
[JsonProperty(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
```

PR #536, opened 2017-12-03
Fix issue with chat member admin tests

Tuscen:
Some of chat member administration tests fail when a regular user has `_` (underscore) in his/her username because underscore is a valid markdown token. Test notification can’t be sent because it has `ParseMode` set as `Markdown` and Telegram can’t find closing underscore and return parsing error that becomes an unexpected `ApiResponseException`.


pouladpld:
We should have just escaped the underscore.

```c#
name.Replace("_", @"\_")
```

PR #550, opened 2017-12-07
Add currency and prices test configuration

Tuscen:
This PR adds configuration for currency and a list of prices in payments tests because not every payments provider accepts every currency and every currency has a minimum total sum.

### DIFF
--- a/test/Telegram.Bot.Tests.Integ/appsettings.json
+++ b/test/Telegram.Bot.Tests.Integ/appsettings.json
@@ -7,6 +7,8 @@
 
   "PaymentProviderToken": "",
   "TesterPrivateChatId": "",
+  "Currency": "",
+  "Prices": "",
 
   "StickerOwnerUserId": "",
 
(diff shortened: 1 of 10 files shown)


test/Telegram.Bot.Tests.Integ/appsettings.json
```diff
@@ -7,10 +7,12 @@
 
   "PaymentProviderToken": "",
   "TesterPrivateChatId": "",
+  "Currency": "",
+  "Prices": "",
```
pouladpld: These new settings should also be documented in sys integ test wikis.

Tuscen:
I'll look into it. Probably you're right. It's easier just to generate a list of prices based on minimum and maximum amount of money.

test/Telegram.Bot.Tests.Integ/appsettings.json
Tuscen: Okay, will document them.

PR #552, opened 2017-12-11
Run tests with dotnet-xunit

karb0f0s:
Run tests with dotnet xunit allows to stop tests run after test fail


karb0f0s:
There is a few caveats with dotnet xunit:
1) You have to reference <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.1" /> in .csproj
2) dotnet xunit works only in the current project context - you have to switch working dir (cd test/...)
3) dotnet xunit does not work until you specify <RuntimeFrameworkVersion>2.0.0</RuntimeFrameworkVersion> (details here: [The specified framework 'Microsoft.NETCore.App', version '2.0' was not found](https://github.com/dotnet/cli/issues/7901) )

Issue #556, opened 2017-12-13
Use overloads for inline versions of some methods instead of using Inline word in their names


Tuscen:
Right now some methods with inline versions use overloads and some use methods with word `Inline` in them. This makes naming inconsistent. I propose to use overloads for all of them.

_NuGet Package Version:_ 13.4.0

Issue #555, opened 2017-12-13
Overload of SetGameScore with inline_message_id returns Message instead of bool


Tuscen:
_NuGet Package Version:_ 13.4.0

Issue #554, opened 2017-12-13
Both SetGameScoreAsync methods have unnecessary parameter editMessage


Tuscen:
[`setGameScore `](https://core.telegram.org/bots/api#setGameScore) doesn't have parameter with that name.

_NuGet Package Version:_ 13.4.0

Issue #544, opened 2017-12-04
SendGameAsync should accept chat_id only as integer and not as ChatId


Tuscen:
`ITelegramBotClient.SendGameAsync` should accept `chat_id` only as `integer` and not as `ChatId`. [`sendGame`](https://core.telegram.org/bots/api#sendgame) accepts only integers as `chat_id`. A user can pass wrong value.

_NuGet Package Version:_ 13.4.0-rc2

PR #557, opened 2017-12-13
Add request classes and fix bugs

Tuscen:
This PR adds request classes and fixes several bugs such as:

- Overload of `SetGameScore` with `inlineMessageId` returns `bool` instead of `Message` (#555)
- `SetWebhookAsync` returns `true` on success [#525]
- `GetGameHighScoresAsync` now accepts `chatId` as `long` instead of `ChatId` (#554)
- `SendGameAsync` now accepts `chatId` as `long` instead of `ChatId` (#544)
- `SendInvoiceAsync` now accepts `chatId` as `long` instead of `ChatId` (#542)
- `SendGameAsync` now accepts `replyMarkup` as `InlineKeyboardMarkup` instead of `IReplyMarkup` (#543 )
- `EditMessageLiveLocationAsync` now accepts `replyMarkup` as `InlineKeyboardMarkup` instead of `IReplyMarkup`
- `EditMessageReplyMarkupAsync` now accepts `replyMarkup` as `InlineKeyboardMarkup` instead of `IReplyMarkup` (#533)
- `EditMessageCaptionAsync` now accepts `replyMarkup` as `InlineKeyboardMarkup` instead of `IReplyMarkup` (#531 and #532)
- `EditMessageTextAsync` now accepts `replyMarkup` as `InlineKeyboardMarkup` instead of `IReplyMarkup` (#530)
- `StopMessageLiveLocationAsync` now accepts `replyMarkup` as `InlineKeyboardMarkup` instead of `IReplyMarkup` (#529)
- `UploadStickerFileAsync` now accepts `pngSticker` as `Stream` instead of `FileToSend` object
- `SetChatPhotoAsync` now accepts `photo` as `Stream` instead of `FileToSend` object (#527)
- `SetWebhookAsync` now accepts `certificate` as `Stream` instead of `FileToSend` object (#526)
- `CreateNewStickerSetAsync` is now spelled correctly (#538)

This PR also make a few changes in `ITelegramBotClient` interface:

- Inline versions of some methods replaced with overloads to ensure naming consistency (#556)
- Some default values of request parameters that are default on Bot API side replaced with `default` keyword and aren't sent with their default values because it doesn't change the behavior of these methods
- Parameter type `T[]` in request methods changed to `IEnumerable<T>`
- Removed redundant return types on methods where the return type on success is always `true` or otherwise the method throws `ApiResponseException`


Tuscen:
Though not all request classes have tests as of now. In particular games related methods, all methods dealing with webhooks, a few editing inline requests and a few others.

Olfi01:
Wow, nice work you did there! I'll be happy to look over it if I get some time, though I'm kinda busy rn 😅

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -41,7 +42,11 @@ public class TelegramBotClient : ITelegramBotClient
 
         private static readonly JsonSerializerSettings SerializerSettings = new JsonSerializerSettings
         {
-            NullValueHandling = NullValueHandling.Ignore
+            NullValueHandling = NullValueHandling.Ignore,
+            ContractResolver = new DefaultContractResolver
```
pouladpld: I know request serialization are only in library but I'd prefer to create a set of reusable request classes that could potentially be used outside of Bot API and don't rely on specific serialization settings to be provided.

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -42,7 +40,11 @@ public class TelegramBotClient : ITelegramBotClient
...
```
pouladpld: I know request serialization are only in library but I'd prefer to create a set of reusable request classes that could potentially be used outside of Bot API and don't rely on specific serialization settings to be provided.

Tuscen: Could you expand on it a little more? I don't understand what you mean by "outside of Bot API".

pouladpld: Some of our request classes could be used in other tg apps such as a tg client.

Also, users now can log/serialize requests before sending them so it is necessary to have all required serialization attributes on classes. No need to require users to pass serialixation settings.

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -408,21 +408,21 @@ public void StopReceiving()
...
+                Offset = offset,
+                Limit = limit,
+                Timeout = timeout
             };
 
-            if (allowedUpdates != null && !allowedUpdates.Contains(UpdateType.All))
-                parameters.Add("allowed_updates", allowedUpdates);
-
-            return SendWebRequestAsync<Update[]>("getUpdates", parameters, cancellationToken);
+            return await MakeRequestAsync(getUpdateRequest, cancellationToken);
```
pouladpld: async/await result in  more code (state machine) to be generated by compiler. we don't need that here. 

such methods could just be an expression like many others you already implemented.

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -457,24 +457,20 @@ public void StopReceiving()
...
             };
 
-            if (allowedUpdates != null && !allowedUpdates.Contains(UpdateType.All))
-                parameters.Add("allowed_updates", allowedUpdates);
-
-            if (certificate != null)
-                parameters.Add("certificate", certificate);
-
-            return SendWebRequestAsync<bool>("setWebhook", parameters, cancellationToken);
+            return await MakeRequestAsync(setWebhookRequest, cancellationToken);
```
pouladpld: This also should be a simple `Task`-returning expression.

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -457,24 +457,20 @@ public void StopReceiving()
         /// If you're having any trouble setting up webhooks, please check out this <see href="https://core.telegram.org/bots/webhooks">amazing guide to Webhooks</see>.
         /// </remarks>
         /// <see href="https://core.telegram.org/bots/api#setwebhook"/>
-        public Task SetWebhookAsync(string url = "", FileToSend? certificate = null,
-            int maxConnections = 40,
-            UpdateType[] allowedUpdates = null,
+        public async Task<bool> SetWebhookAsync(
+            string url = "",
```
pouladpld: This tg API methods is a bit confusing. passing empty url here is exatcly the same as calling `deleteWebhook`!

We shouldn't have default value for `url` param. Users can either pass empty here or call `DeleteWebhookAsync` to remove the webhook.

pouladpld: Also, if we do such as change, let's have it on the migration wiki.

pouladpld:
Almost all auto-implemented properties in request and response classes could be replaced with fields. No need generate more redundant code for properties.

```c#
public string Url { get; set; }
// turns into:
public string Url;
```

src/Telegram.Bot/Requests/SetWebhookRequest.cs
```diff
@@ -0,0 +1,115 @@
...
+        { }
+
+        /// <summary>
+        /// Generate content of HTTP message
+        /// </summary>
+        /// <param name="serializerSettings">JSON serialization setting</param>
+        /// <returns>Content of HTTP request</returns>
+        public override HttpContent ToHttpContent(JsonSerializerSettings serializerSettings)
+        {
+            var multipartContent = new MultipartFormDataContent(Guid.NewGuid().ToString() + DateTime.UtcNow.Ticks)
```
pouladpld: I'm not really sure whether generating a new GUID every single time for mutlipart requests is a good idea. I started doing it before but we might want to reassess this approach.

pouladpld:
The logic for creating a `multi-part/form` http content is duplicated in many places. Can we have a general method for it in `RequestBase` class?

However, implementing such a general method might result in a performance hit.

src/Telegram.Bot/Requests/SetWebhookRequest.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// Specify a url and receive incoming updates via an outgoing webhook
+    /// </summary>
+    public class SetWebhookRequest : RequestBase<bool>
+    {
+        private IEnumerable<UpdateType> _allowedUpdateTypes;
+
+        /// <summary>
+        /// HTTPS url to send updates to. Use an empty string to remove webhook integration.
+        /// </summary>
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
```
pouladpld: This request won't be JSON serialized so no need to use any JSON attribute.

src/Telegram.Bot/Requests/SetWebhookRequest.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <summary>
+        /// Generate content of HTTP message
+        /// </summary>
+        /// <param name="serializerSettings">JSON serialization setting</param>
+        /// <returns>Content of HTTP request</returns>
+        public override HttpContent ToHttpContent(JsonSerializerSettings serializerSettings)
+        {
+            var multipartContent = new MultipartFormDataContent(Guid.NewGuid().ToString() + DateTime.UtcNow.Ticks)
+            {
+                { new StringContent(Url), nameof(Url).ToSnakeCased() }
```
pouladpld: Name of request params are very unlikely to change so we can gain a better performance by just hard coding them.

```c#
nameof(Url).ToSnakeCased()
// turns into:
"url"
```

src/Telegram.Bot/Requests/SetWebhookRequest.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// Public key certificate so that the root certificate in use can be checked
+        /// </summary>
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public Stream Certificate { get; set; }
+
+        /// <summary>
+        /// Maximum allowed number of simultaneous HTTPS connections to the webhook for update delivery
+        /// </summary>
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public int MaxConnections { get; set; }
```
pouladpld: Values are _1-100. Defaults to 40_. We could statically enforce the range better by using `uint` instead of `int`.

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -479,8 +479,8 @@ public void StopReceiving()
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
         /// <returns>Returns true on success</returns>
         /// <see href="https://core.telegram.org/bots/api#deletewebhook"/>
-        public Task<bool> DeleteWebhookAsync(CancellationToken cancellationToken = default)
-            => SendWebRequestAsync<bool>("deleteWebhook", null, cancellationToken);
+        public async Task<bool> DeleteWebhookAsync(CancellationToken cancellationToken = default)
+            => await MakeRequestAsync(new DeleteWebhookRequest(), cancellationToken);
```
pouladpld: No need to await.

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -479,8 +479,8 @@ public void StopReceiving()
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
         /// <returns>Returns true on success</returns>
         /// <see href="https://core.telegram.org/bots/api#deletewebhook"/>
-        public Task<bool> DeleteWebhookAsync(CancellationToken cancellationToken = default)
-            => SendWebRequestAsync<bool>("deleteWebhook", null, cancellationToken);
+        public async Task<bool> DeleteWebhookAsync(CancellationToken cancellationToken = default)
```
pouladpld: No need to return anything. If there are any errors, an exception would be thrown so return value(if successful) would always be `true` which is unnecessary to care for.

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -488,8 +488,8 @@ public void StopReceiving()
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
         /// <returns>On success, returns <see cref="WebhookInfo"/>.</returns>
         /// <see href="https://core.telegram.org/bots/api#getwebhookinfo"/>
-        public Task<WebhookInfo> GetWebhookInfoAsync(CancellationToken cancellationToken = default)
-            => SendWebRequestAsync<WebhookInfo>("getWebhookInfo", null, cancellationToken);
+        public async Task<WebhookInfo> GetWebhookInfoAsync(CancellationToken cancellationToken = default)
+            => await MakeRequestAsync(new GetWebhookInfoRequest(), cancellationToken);
```
pouladpld: Change to Task-returning expression

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -501,8 +501,8 @@ public void StopReceiving()
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
         /// <returns>Returns basic information about the bot in form of <see cref="User"/> object</returns>
         /// <see href="https://core.telegram.org/bots/api#getme"/>
-        public Task<User> GetMeAsync(CancellationToken cancellationToken = default)
-            => SendWebRequestAsync<User>("getMe", cancellationToken: cancellationToken);
+        public async Task<User> GetMeAsync(CancellationToken cancellationToken = default)
+            => await MakeRequestAsync(new GetMeRequest(), cancellationToken);
```
pouladpld: should be a Task-returning expression

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -517,24 +517,26 @@ public void StopReceiving()
...
+                DisableWebPagePreview = disableWebPagePreview,
+                DisableNotification = disableNotification,
+                ReplyToMessageId = replyToMessageId,
+                ReplyMarkup = replyMarkup
+            };
 
-            return SendMessageAsync(MessageType.TextMessage, chatId, text, disableNotification, replyToMessageId,
-                replyMarkup,
-                additionalParameters, cancellationToken);
+            return await MakeRequestAsync(request, cancellationToken);
```
pouladpld: should be a Task-returning expression

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -549,21 +549,19 @@ public void StopReceiving()
...
-                {"from_chat_id", fromChatId},
-                {"message_id", messageId},
+                DisableNotification = disableNotification
             };
 
-            if (disableNotification)
-                parameters.Add("disable_notification", true);
-
-            return SendWebRequestAsync<Message>("forwardMessage", parameters, cancellationToken);
+            return await MakeRequestAsync(request, cancellationToken);
```
pouladpld: should be a Task-returning expression

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -576,19 +576,24 @@ public void StopReceiving()
...
-                {"caption", caption}
+                Caption = caption,
+                ReplyToMessageId = replyToMessageId,
+                DisableNotification = disableNotification,
+                ReplyMarkup = replyMarkup
             };
 
-            return SendMessageAsync(MessageType.PhotoMessage, chatId, photo, disableNotification, replyToMessageId,
-                replyMarkup, additionalParameters, cancellationToken);
+            return await MakeRequestAsync(request, cancellationToken);
```
pouladpld: should be a Task-returning expression

src/Telegram.Bot/Requests/SetWebhookRequest.cs
Tuscen: We can use predefined string to be used as boundary in every request with multipart/form-data.

Tuscen:
I think we could do ourselves a disservice if we will use fields instead of properties in public facing code. Fields can be passed as ref or out parameters whereas properties can not. If someone's code is depended on fields' behavior and then we change the field to a property it will be a breaking change.

src/Telegram.Bot/TelegramBotClient.cs
Tuscen: Yeah, valid point about logging requests.

src/Telegram.Bot/Requests/SetWebhookRequest.cs
Tuscen: Oh, I forgot about this constraint. It can be implemented in the getter. So all values less than 1 or greater than 100 would not be sent. Though we can't enforce this in `ITelegramBotClient.SetWebhookAsync` and it would be possible to pass any valid number to it.

src/Telegram.Bot/TelegramBotClient.cs
Tuscen: Agree, I don't like it. I'm for removing default url.

src/Telegram.Bot/TelegramBotClient.cs
Tuscen: Though not all requests can be serialized to json. E.g. requests with files. You also commented on `SetWebhookRequest.cs` that json attributes can be ommited because it is always use  `multipart/form-data`. How can you log such requests?

pouladpld: We could add an extension method for that.
```c#
// Generate HTTP Content and then stringify it
string logMe = reqeust.ToPayloadString()
```
this method would also be used in `SendingRequest` and `ResponseReceived` events

src/Telegram.Bot/Requests/SetWebhookRequest.cs
pouladpld: having a constant value is a good idea but what unique string could we use that isn't in another part of the payload?

src/Telegram.Bot/Requests/SetWebhookRequest.cs
pouladpld: Request validation is another nice feature we can implement. We'll look into it later. For now, just a simple type change would help a bit.

src/Telegram.Bot/Requests/SetWebhookRequest.cs
Tuscen: I guess any random string would be sufficient. We can generate GUID somewhere inside internal class in a static constructor so it will be unique on every application start. How high is the chance of a collision with something inside of the payload during single app lifetime cycle?

src/Telegram.Bot/Requests/SendPhotoRequest.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// </summary>
+        /// <param name="serializerSettings">JSON serialization setting</param>
+        /// <returns>Content of HTTP request</returns>
+        public override HttpContent ToHttpContent(JsonSerializerSettings serializerSettings)
+        {
+            HttpContent content;
+
+            if (Photo.Type == FileType.Stream)
+            {
+                var parameters = new Dictionary<string, object>
```
pouladpld: > Side note. Is there a way to avoid using dictionaries?

src/Telegram.Bot/Requests/SendPhotoRequest.cs
```diff
@@ -0,0 +1,108 @@
...
+                    { nameof(ChatId).ToSnakeCased(), ChatId},
+                    { nameof(Photo).ToSnakeCased(), Photo },
+                    { nameof(Caption).ToSnakeCased(), Caption },
+                    { nameof(ReplyMarkup).ToSnakeCased(), ReplyMarkup }
+                };
+
+                if (ReplyToMessageId != default)
+                {
+                    parameters.Add(nameof(ReplyToMessageId).ToSnakeCased(), ReplyToMessageId);
+                }
```
pouladpld: > Side note. Get the content for these using extension methods on types like `INotifiableMessage` and `IReplyMessage`

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -611,23 +611,30 @@ public void StopReceiving()
...
+                Performer = performer,
+                Title = title,
+                DisableNotification = disableNotification,
+                ReplyToMessageId = replyToMessageId,
+                ReplyMarkup = replyMarkup
             };
 
-            return SendMessageAsync(MessageType.AudioMessage, chatId, audio, disableNotification, replyToMessageId,
-                replyMarkup, additionalParameters, cancellationToken);
+            return await MakeRequestAsync(request, cancellationToken);
```
pouladpld: no await

src/Telegram.Bot/Requests/SetWebhookRequest.cs
Tuscen: Though constant value even during single app cycle could lead to greater chance of a collision. I think generating the boundary on every request is not that bad. What do you think is more performant: using Guid or generating random string using preallocated StringBuilder that is cleared after every ToString call?

Something like:
```
builder.Clear();
var rnd = new Random();
...
builder.Append((char)(rnd.Next(65, 90)))
return builder.ToString();
```

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -650,20 +650,24 @@ public void StopReceiving()
...
+                Caption = caption,
+                DisableNotification = disableNotification,
+                ReplyToMessageId = replyToMessageId,
+                ReplyMarkup = replyMarkup
             };
 
-            return SendMessageAsync(MessageType.DocumentMessage, chatId, document, disableNotification,
-                replyToMessageId,
-                replyMarkup, additionalParameters, cancellationToken);
+            return await MakeRequestAsync(request, cancellationToken);
```
pouladpld: no await

src/Telegram.Bot/Requests/FileRequestBase.cs
```diff
@@ -0,0 +1,80 @@
...
+        { }
+
+        /// <summary>
+        /// Initializes an instance of request
+        /// </summary>
+        /// <param name="methodName">Bot API method</param>
+        /// <param name="method">HTTP method to use</param>
+        protected FileRequestBase(string methodName, HttpMethod method)
+            : base(methodName, method)
+        { }
```
pouladpld: Not sure whether this ctor will be needed. AFAIK, all file sending requests are POST

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -406,22 +392,21 @@ public void StopReceiving()
...
-                parameters.Add("allowed_updates", allowedUpdates);
-
-            return SendWebRequestAsync<Update[]>("getUpdates", parameters, cancellationToken);
-        }
+                Offset = offset,
+                Limit = limit,
+                Timeout = timeout,
+                AllowedUpdates = allowedUpdates
+            }, cancellationToken)
+            .ContinueWith(t => t.Result.ToArray());
```
Tuscen: Do you remember why we decided to use `async`/`await` instead of `ContinuteWith`? `ContinueWith` throws `AggregateException` with `ApiResponseException` inside `InnerExceptions` collection so we have to unwrap it ourselves and rethrow. With `async`/`await` we can be sure that the compiler generates the necessary code.

pouladpld: That's right. In fact, we should just return the `List<T>` itself. I'll fix it.

src/Telegram.Bot/Requests/AnswerInlineQueryRequest.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            get => _nextOffset;
+            set
+            {
+                if (!string.IsNullOrWhiteSpace(value))
+                {
+                    _nextOffset = value;
+                }
+            }
+        }
```
pouladpld: Why not auto-implemented?

CHANGELOG.md
```diff
@@ -104,17 +102,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Type of parameter `replyMarkup` in method `SendInvoiceAsync` changed to `InlineKeyboardMarkup`
 - Type of parameter `prices` in method `SendInvoiceAsync` changed to `IEnumerable<LabeledPrice>`
 - Type of parameter `shippingOptions` in method `AnswerShippingQueryAsync` changed to `IEnumerable<ShippingOption>`
-- Type of parameter `chatId` in method `SendInvoiceAsync` changed to `long`
+- Type of parameter `chatId` in method `SendInvoiceAsync` changed to `int`
```
arhont96: Why, actually? As far as I understand, chatId must be long, not int.

tuscen: Invoices are always sent in private chats. And private chat ids are always in range of 32 bits.

PR #583, opened 2018-01-12
Use reflection to find RunSummary property

karb0f0s:
Use reflection on fixture class instead of explicit casting to TestsFixture class. This would allow to have multiple Assembly Fixtures and to not rely on RunSummary property naming


pouladpld:
@karb0f0s my initial thought was to avoid multiple assembly fixtures. However, some props could be separated from `TestsFixture` such as `ChannelChat`, `PrivateChat`, and `AllowedUserNames`.

karb0f0s:
Well, with this kind of tight coupling there is no need to use AssemblyFixtureAttribute and it is possible to simplify AfterTestAssemblyStartingAsync to this:
```c#
            Aggregator.Run(() =>
                _assemblyFixtureMappings.Add(typeof(TestsFixture), Activator.CreateInstance(typeof(TestsFixture))));
```

PR #584, opened 2018-01-12
Make parameterless constructors for InlineQueryResults private and other changes

Banane9:
* Made the parameterless constructors for InlineQueryResults private (Json.Net can still use them)
* Made required properties for the MessageContent objects have private setters and added constructors to them
* Made the StopReceiving() method on the BotClient not throw WebExceptions or TaskCanceled exceptions


Banane9:
Added custom equality based on a User's Id (instead of default reference equality) to the User class, so that you can use the user object as key in a dictionary for example.

pouladpld:
@Banane9 I pushed a few changes here
<https://github.com/TelegramBots/telegram.bot/commits/pr584>

PR #587, opened 2018-01-16
pr584

pouladpld:
Following #584 by @Banane9

### DIFF
--- a/src/Telegram.Bot/Types/User.cs
+++ b/src/Telegram.Bot/Types/User.cs
@@ -45,24 +45,39 @@ namespace Telegram.Bot.Types
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string LanguageCode { get; set; }
 
-        /// <inheritdoc/>
+        /// <inheritdoc />
         public override bool Equals(object obj)
         {
-            var userObj = obj as User;
+            if (!(obj is User other))
+                return false;
 
-            return userObj != null && userObj.Id == Id;
+            return Id == other.Id &&
+                   IsBot == other.IsBot &&
+                   FirstName == other.FirstName &&
+                   LastName == other.LastName &&
+                   Username == other.Username &&
+                   LanguageCode == other.LanguageCode;
         }
 
-        /// <inheritdoc/>
+        /// <inheritdoc />
         public override int GetHashCode()
         {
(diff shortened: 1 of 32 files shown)


test/Telegram.Bot.Tests.Unit/UserTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            User user1Copy = JsonConvert.DeserializeObject<User>(json1);
+            User user2 = JsonConvert.DeserializeObject<User>(json2);
+
+            Dictionary<User, string> dict = new Dictionary<User, string>
+            {
+                {user1, nameof(user1)},
+                {user2, nameof(user2)}
+            };
+
+            Assert.Equal(nameof(user1), dict[user1]);
```
Banane9: This would be true regardless of implementing custom equality; use user1Copy for accessing the dictionary to test custom equality

src/Telegram.Bot/Types/User.cs
```diff
@@ -44,5 +45,60 @@ public class User
...
+        /// <inheritdoc />
+        public override bool Equals(object obj) => Equals(obj as User);
+
+        /// <inheritdoc />
+        public bool Equals(User other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
+
+            return Id == other.Id &&
```
Banane9: Specifically only used Id to compare, so that a user still compares the same, if they change their name at any time during conversation with the bot

src/Telegram.Bot/Types/User.cs
```diff
@@ -44,5 +45,60 @@ public class User
...
+        /// <summary>
+        /// Compares two users for inequality
+        /// </summary>
+        /// <param name="lhs">Left-hand side user in expression</param>
+        /// <param name="rhs">Right-hand side user in expression</param>
+        /// <returns>true if users are not equal, otherwise false</returns>
+        public static bool operator !=(User lhs, User rhs) => !(lhs == rhs);
+
+        /// <inheritdoc />
+        public override int GetHashCode()
```
Banane9: hashcode should reflect the equality, if that's changed

Banane9:
Just noticed this now, somehow - thanks for merging!

PR #598, opened 2018-02-10
Fix descriptions of two properties of InlineKeyboardButton

Olgabrezel:


pouladpld:
@Olgabrezel Thank you

PR #605, opened 2018-02-24
ParseMode to Caption video for InlineQueryResultVideo

Startname:
Hello
Please add parse_mode to caption video for InlineQueryResultVideo


pouladpld:
@Startname We will release soon.

PR #608, opened 2018-02-24
Add parse_mode field to inline query results

Tuscen:


pouladpld:
Can we address issue #610 in this PR as well?

PR #631, opened 2018-03-06
Fix error CS8363 A default literal 'default' is not valid as a pattern

karb0f0s:
- Fix error CS8363 as a result of proposal [Disallow `default` as a constant pattern](https://github.com/dotnet/roslyn/issues/23499)


pouladpld:
You closed it. Is the issue resolved?

karb0f0s:
@Tuscen said it's better to create PRs to devel branch. I'll make new PR to dev a bit later

PR #632, opened 2018-03-06
Clarify User.ToString method

karb0f0s:
It is not obvious how full username is generated. I suggest to use local function to clarify username creation


pouladpld:
I don't really see a need for this PR.

Issue #640, opened 2018-03-10
EditorConfig


poulad:
There are various dev envs that we use. In order to have consistency through our code, I suggest using [EditorConfig](http://editorconfig.org/).

@Tuscen suggests <https://github.com/RehanSaeed/EditorConfig/blob/master/.editorconfig>

Tuscen:
I suggest to use it as a base and tweak it.

In particular:
- Use LF for everything instead of CRLF
- Insert final new line
- Change `dotnet_style_qualification_for_*` to `false`
- Change `csharp_preserve_single_line_blocks` to `false`
- Add `csharp_prefer_simple_default_expression = true`
- Add `csharp_prefer_braces  = true`

Everything else seems fine.

See https://kent-boogaart.com/blog/editorconfig-reference-for-c-developers

karb0f0s:
As @pouladpld mentioned once - tests require explicit type specification instead of `var` declaration. So it should be separate configuration for these cases.

pouladpld:
@Tuscen I believe we should mark them as errors. However, CI checks will happen on `develop` branch so a bit of messy code on feature branches(before getting to `develop`) would be fine.

pouladpld:
@Tuscen suggests StyleCop as an alternative.

Tuscen:
@pouladpld I guess we still need editorconfig to have common editor configs like line ending and such.

pouladpld:
@Tuscen well, we can have more than 1

karb0f0s:
could be a good example of codestyle
https://github.com/TelegramBots/Telegram.Bot/blob/ff61b89527f493ee081a2161ff304d4941357d73/src/Telegram.Bot/Exceptions/ForbiddenExceptions/ChatNotInitiatedException.cs#L13
vs
https://github.com/TelegramBots/Telegram.Bot/blob/ff61b89527f493ee081a2161ff304d4941357d73/src/Telegram.Bot/Exceptions/BadRequestExceptions/BotBlockedException.cs#L12
single line vs multiline. braces on same line vs braces on single line

wiz0u:
No need for such discussion if there is mostly only one main programmer now. He will decide his coding style.

PR #652, opened 2018-03-18
Add editorconfig

Tuscen:
This PR will add `.editorconfig` file to prevent constant style changes in files between commits from different collaborators.

Following default configs are set:
- Charset is UTF-8
- Use LF for line endings
- Insert final new line
- Use spaces for indentation
- Indentation size is 4 spaces
- Trim trailing whitespace

Trimming trailing whitespace is turned off for markdown because it can break the markup.

Force explicit type instead of 'var' in integration tests project
- add .editorconfig to force explicit type instead of 'var' in integration tests project
- allow implicit type declaration for tests framework

#640

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,29 @@
+# http://EditorConfig.org
+
+# This file is the top-most EditorConfig file
+root = true
+
+# All Files
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# Solution Files
+[*.sln]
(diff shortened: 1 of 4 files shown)


.editorconfig
```diff
@@ -0,0 +1,29 @@
...
+
+# This file is the top-most EditorConfig file
+root = true
+
+# All Files
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+indent_size = 4
```
pouladpld: How about using tabs with size of 2?

karb0f0s: As of now all files have tab size 4, so all of them will be changed after reformatting

Issue #648, opened 2018-03-14
Add "Forbidden: bot is not a member of the channel chat" exception


tuscen:
HTTP code: `403`
Error message: ` Forbidden: bot is not a member of the channel chat`

karb0f0s:
How do you get this? Send message to some public channel when you are not its member?

Tuscen:
Either public or private, doesn't matter.

karb0f0s:
I was able to test this exception in supergroup and channel, it is not thrown for regular group - ChatNotFoundException is thrown instead

karb0f0s:
[Add BotIsNotMemberException](https://github.com/TelegramBots/Telegram.Bot/commit/0dd29b39496e3e88704932564af5cc69437030ca)

karb0f0s:
anyone can think of better a name?

Tuscen:
It should somehow include word "channel" because it is thrown only for channels.

pouladpld:
If such errors happen for other chats, we could use general base and derived subclasses.

karb0f0s:
@Tuscen actually not 'bot is not a member of the (supergroup|channel) chat'

Tuscen:
@karb0f0s yeah, makes sense to have single type if it can be thrown with slightly different error message for other chats.

karb0f0s:
This issue was moved to TelegramBots/Telegram.Bot.Extensions.Exceptions#3

AngryBandera:
> This issue was moved to [TelegramBots/Telegram.Bot.Extensions.Exceptions#3](https://github.com/TelegramBots/Telegram.Bot.Extensions.Exceptions/issues/3)

Have you fixed this issue? I have bot, but one day he started throwing this error. I idk what to do.

Issue #646, opened 2018-03-12
Clean ApiException.cs


karb0f0s:
We should clean this commented out part in 
[public static ApiRequestException FromApiResponse<T>(ApiResponse<T> apiResponse)](https://github.com/TelegramBots/Telegram.Bot/blob/d94d7e6f295e40eb1da504b38569a7c762c60807/src/Telegram.Bot/Exceptions/ApiException.cs#L85-L161)
some exceptions already implemented, and there is no need to keep dead code for them, some not implemented yet, like [case "bot was blocked by the user"](https://github.com/TelegramBots/Telegram.Bot/blob/d94d7e6f295e40eb1da504b38569a7c762c60807/src/Telegram.Bot/Exceptions/ApiException.cs#L137) but declared, which is misleading

karb0f0s:
Should `BadRequestException` and `ForbiddenException` abstract classes be renamed to `BadRequestExceptionBase` and `ForbiddenExceptionBase`?
Could then `BadRequestException` and `ForbiddenException` classes be used to throw general 400 and 403 errors?

Tuscen:
@karb0f0s Agree, it makes sense to throw general `BadRequestException` and `ForbiddenException`  when there is no particular exception type.

karb0f0s:
https://github.com/TelegramBots/Telegram.Bot/pull/653

karb0f0s:
Closed due to inactivity

Issue #628, opened 2018-03-05
Add exception type for too many requests


tuscen:
HTTP Code: `429`
Error message: `Too Many Requests: retry after [0-9]+`

pouladpld:
What's HTTP status code?

karb0f0s:
429?

Tuscen:
@karb0f0s I added it after @pouladpld had asked :) Forgot about status code when I opened the issue.

pouladpld:
![image](https://user-images.githubusercontent.com/14295804/38101161-87287c00-334d-11e8-8174-7bbe9791cff9.png)

karb0f0s:
https://github.com/TelegramBots/Telegram.Bot/pull/657/commits/48fe2a57f0783f5cdb42c7b356d1bb89b1cdb4b9

karb0f0s:
wow https://github.com/dotnet/corefx/issues/4382

karb0f0s:
BTW I think constants should be replaced with HttpStatusCode - either exception class accept HttpStatusCode, or error code will be cast to int?
https://github.com/TelegramBots/Telegram.Bot/blob/ce94da88aecb870b7e7a263d2f0e50da82beaba0/src/Telegram.Bot/TelegramBotClient.cs#L222-L225
https://github.com/TelegramBots/Telegram.Bot/blob/ce94da88aecb870b7e7a263d2f0e50da82beaba0/src/Telegram.Bot/TelegramBotClient.cs#L243-L245
https://github.com/TelegramBots/Telegram.Bot/blob/ce94da88aecb870b7e7a263d2f0e50da82beaba0/src/Telegram.Bot/TelegramBotClient.cs#L198
https://github.com/TelegramBots/Telegram.Bot/blob/ce94da88aecb870b7e7a263d2f0e50da82beaba0/src/Telegram.Bot/TelegramBotClient.cs#L297

Maybe the same should be made here
https://github.com/TelegramBots/Telegram.Bot/blob/ce94da88aecb870b7e7a263d2f0e50da82beaba0/src/Telegram.Bot/Exceptions/BadRequestException.cs#L10
and here
https://github.com/TelegramBots/Telegram.Bot/blob/ce94da88aecb870b7e7a263d2f0e50da82beaba0/src/Telegram.Bot/Exceptions/ForbiddenException.cs#L10
bcs at first there is a check for HttpStatusCode and after comes some constants

matteocontrini:
[...]

PR #657, opened 2018-03-28
Exceptions

karb0f0s:
## Refactor `ApiExceptionParser.Parse`
- `BadRequestException` is thrown on `Error 400: Bad request` if more specific exception is not defined
- `ForbiddenException` is thrown on `Error 403: Forbidden` if more specific exception is not defined
- add optional parameters to `InvalidParameterException` derived classes
- fix tests to handle `BadRequestException`
- separate collections for different exception types

## Clean `ApiException` class
- remove commented out exception code

## Bring back exceptions
- exception `BotBlockedException`
- exception `BotRestrictedException`
- exception `MissingParameterException`
- exception `NotEnoughRightsException`
- exception `WrongChatTypeException`
- add test case for `MissingParameterException`
- add test cases for `WrongChatTypeException`
- fix `Should_Throw_On_Setting_Chat_Sticker_Set` test to handle `WrongChatTypeException`

## Changed
- arrange exception classes in folders according to their base classes

## Added
- exception `ChatDescriptionIsNotModifiedException`
- exception `ChatNotModifiedException`
- exception `SendMessageToBotException`
- exception `TooManyRequestsException`
- exception `BotIsNotMemberException`

#646 #628 #648


pouladpld:
@karb0f0s is there any breaking change? If not, we will release new exceptions later.

Issue #642, opened 2018-03-11
Send test case notification via framework


pouladpld:
We can have test case description set only once on the `OrderedFact`. Then, test notification message would be sent to chat via the framework prior to running the test method.

> Not sure how to do that as of now.

```c#
[OrderedFact(DisplayName = "should do something")]
public Task Should_Do(Message testCaseNotificationMessage) // Optional arg
{ /* ... */ }
```

karb0f0s:
[XunitTestInvoker.cs](https://github.com/xunit/xunit/blob/master/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestInvoker.cs) contains methods [BeforeTestMethodInvokedAsync](https://github.com/xunit/xunit/blob/94ae1736405f22b86d274245d11c54c22b7d680d/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestInvoker.cs#L51) and [InvokeTestMethodAsync](https://github.com/xunit/xunit/blob/94ae1736405f22b86d274245d11c54c22b7d680d/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestInvoker.cs#L103) that allow do such things. but I do not have any idea how to make framework use inherited class

karb0f0s:
done. also we could announce which tests collection is running

karb0f0s:
@pouladpld 
With #666 merged it is possible to extend [`OrderedFactAttribute`](https://github.com/TelegramBots/Telegram.Bot/blob/develop/test/Telegram.Bot.Tests.Integ/Framework/OrderedFactAttribute.cs) with `Instructions`, `StartInlineQuery` parameters, that will reflect parameters for 
```csharp
SendTestCaseNotificationAsync(string testcase,
            string instructions = default,
            ChatId chatid = default,
            bool startInlineQuery = default)
```
and replace this method in the test with annotations
```csharp
[OrderedFact(DisplayName = FactTitles.ShouldAnswerInlineQueryWithArticle, 
             Instructions = "Click on location message's button to edit the location",
             StartInlineQuery = true)]
```
instead of 
```csharp
SendTestCaseNotificationAsync(...., startInlineQuery: true);
```

PR #666, opened 2018-04-08
Add custom xUnit class and method runners

karb0f0s:
## Add custom runners to test framework
- add `XunitCustomTestClassRunner`
- add `XunitCustomTestMethodRunner`

This allows to extend test runs: for example add notification (#642) or add delay between test runs.

## Feature: use `Polly` to retry failed tests
- add `Polly` to handle failed tests
- setup `Polly` policy to run test three more times with specified delay if test failed


karb0f0s:
I suggest to close this PR, i found a better solution #691

PR #679, opened 2018-04-30
Update XML documentations

pouladpld:


src/Telegram.Bot/ITelegramBotClient.cs
```diff
@@ -578,21 +578,39 @@ public interface ITelegramBotClient
             string fileId,
             CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Use this method to get download a file. Get <paramref name="filePath"/> by calling <see cref="GetFileAsync"/>
```
tuscen: Should be `Use this method to download a file.`

src/Telegram.Bot/ITelegramBotClient.cs
```diff
@@ -578,21 +578,39 @@ public interface ITelegramBotClient
...
+        /// </summary>
+        /// <param name="filePath">Path to file on server</param>
+        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
+        /// <returns>File stream</returns>
         Task<Stream> DownloadFileAsync(
             string filePath,
             CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Use this method to get download a file. Get <paramref name="filePath"/> by calling <see cref="GetFileAsync"/>
```
tuscen: Same

src/Telegram.Bot/ITelegramBotClient.cs
pouladpld: Fixed

src/Telegram.Bot/ITelegramBotClient.cs
pouladpld: Fixed

Issue #669, opened 2018-04-08
Disable InlineQueryResult caching in Should_Delete_Message_From_InlineQuery


karb0f0s:
https://github.com/TelegramBots/Telegram.Bot/blob/16facadc1c27eaf23ffe8edcdb80382ea037648e/test/Telegram.Bot.Tests.Integ/Update%20Messages/DeleteMessageTests.cs#L52-L61
Add `cacheTime: 0` to AnswerInlineQueryAsync call to avoid result caching
```csharp
await BotClient.AnswerInlineQueryAsync(
    inlineQueryId: queryUpdate.InlineQuery.Id,
    cacheTime: 0,  // <==
    results: new[]
    {
        new InlineQueryResultArticle(
            id: "article-to-delete",
            title: "Telegram Bot API",
            inputMessageContent: new InputTextMessageContent("https://www.telegram.org/")
        )
    }
);
```

pouladpld:
@karb0f0s 
Yes. It's better to have it even though that isn't an issue now. Feel free to make a PR.

karb0f0s:
done https://github.com/TelegramBots/Telegram.Bot/commit/4f88277c4c8f1bff04bfd3fdebd6f3e87e46ce99

Issue #610, opened 2018-02-24
Add asserts to inline query tests


pouladpld:
In inline query tests, ask user to send result to chat and have more assertions for it.

PR #683, opened 2018-05-01
Fix InlineQuery tests

karb0f0s:
- disable result caching in `Should_Delete_Message_From_InlineQuery` #669 
- call `GetInlineQueryResultUpdates` after `AnswerInlineQueryAsync` to check inline query response #610


pouladpld:
Merged (cherry-picked)

karb0f0s:
i’ve done only half of them :(

pouladpld:
My bad. Please go on.

PR #687, opened 2018-05-02
Tests with proxy

karb0f0s:
- add socks5 proxy support in `TestsFixture`

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -6,6 +6,7 @@ branches:
     - docs
 image: Visual Studio 2017
 before_build:
+  - git submodule update --init --recursive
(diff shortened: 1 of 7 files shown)


karb0f0s:
.gitmodules format was wrong and i had to add `- git submodule update --init --recursive` step to pull submodule code from external repository

karb0f0s:
This class is not available as nuget package :(. Other nuget libs I tried do not provide socks5 proxy functionality.

MihaZupan:
The library is now available on NuGet: https://www.nuget.org/packages/HttpToSocks5Proxy/

karb0f0s:
I found a way to run tests without proxy, can I close this PR now?

Issue #684, opened 2018-05-01
Delegates should be called outside try..catch block


karb0f0s:
Delegates in `ReceiveAsync` function should be called outside try..catch block to prevent catching user exceptions
https://github.com/TelegramBots/Telegram.Bot/blob/fb9acf94104ba31772bd3334004c87f3e4b7baef/src/Telegram.Bot/TelegramBotClient.cs#L324-L328

PR #701, opened 2018-05-10
Fix: call delegates outside receive loop

karb0f0s:
- in update receive loop call delegates outside `try..catch` block to avoid catching exceptions in user code

#684


src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -338,6 +334,12 @@ public TelegramBotClient(string token, IWebProxy webProxy)
                 {
                     OnReceiveGeneralError?.Invoke(this, generalException);
                 }
+
+                foreach (var update in updates)
+                {
+                    MessageOffset = update.Id + 1;
```
tuscen: I think update offset increment should be after the event handlers invocation. So if something went wrong and exception was thrown the update is not lost when StartReceiving called again on the same bot client instance.

PR #714, opened 2018-05-16
Separate Interactive and Automated test cases

pouladpld:
- Run interactive tests first
- Divide some test collections into interactive and automated collections

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,12 @@ sudo: false
 mono: none
 dotnet: 2.0.0
 branches:
-  only:
-  - develop
+  except:
+  - master
 script: >
   cd "test/Telegram.Bot.Tests.Integ" &&
   dotnet test --configuration Release --list-tests &&
-  dotnet xunit -configuration Release -nobuild -stoponfail -verbose
+  dotnet xunit -configuration Release -nobuild -stoponfail -verbose $([ "$TRAVIS_BRANCH" != 'develop' ] && echo '-notrait "Category=Interactive"')
 notifications:
   email: false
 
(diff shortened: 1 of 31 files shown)


karb0f0s:
We can further extend idea adding automated integration tests run in appveyor. Unfortunately xUnit test runner have limited options of filtering test cases, but I think we could provide new `[Trait("Category", "Autotest")]` to have this kind of functionality.
```
  -trait "name=value"    : only run tests with matching name/value traits
                         : if specified more than once, acts as an OR operation
  -notrait "name=value"  : do not run tests with matching name/value traits
                         : if specified more than once, acts as an AND operation
  -method "name"         : run a given test method (should be fully specified;
                         : i.e., 'MyNamespace.MyClass.MyTestMethod')
                         : if specified more than once, acts as an OR operation
  -class "name"          : run all methods in a given test class (should be fully
                         : specified; i.e., 'MyNamespace.MyClass')
                         : if specified more than once, acts as an OR operation
  -namespace "name"      : run all methods in a given namespace (i.e.,
                         : 'MyNamespace.MySubNamespace')
```

pouladpld:
@karb0f0s Yes. That's what I'm looking forward. I add trait to interactive tests and ignore them on all branches except `develop`. Sys Integ tests can always be run on Travis CI.

pouladpld:
By adding this trait on _collections only_, we mark test cases interactive.

https://github.com/TelegramBots/Telegram.Bot/blob/83b912bb02143d023b80caa8d97716f34d4153a5/test/Telegram.Bot.Tests.Integ/Admin%20Bot/ChatMemberAdministrationTests.cs#L12

PR #716, opened 2018-05-19
Added better InputMedia constructors

MihaZupan:
Makes it more convenient when sending albums and avoids confusion with InlineMedia* types


src/Telegram.Bot/Types/InputFiles/InputMediaPhoto.cs
```diff
@@ -13,5 +15,19 @@ public InputMediaPhoto()
...
+
+        /// <summary>
+        /// Initializes a new photo media to send with an <see cref="InputMedia"/>
+        /// </summary>
+        public InputMediaPhoto(
+            InputMedia media,
+            string caption = default,
+            ParseMode parseMode = default)
+        {
+            Type = "photo";
```
tuscen: You should just call default ctor here.

src/Telegram.Bot/Types/InputFiles/InputMediaVideo.cs
```diff
@@ -40,5 +41,27 @@ public InputMediaVideo()
...
+        public InputMediaVideo(
+            InputMedia media,
+            int duration = default,
+            int width = default,
+            int height = default,
+            string caption = default,
+            ParseMode parseMode = default,
+            bool supportsStreaming = default)
+        {
+            Type = "video";
```
tuscen: Same

MihaZupan:
Updated and squashed into 1 commit

PR #717, opened 2018-05-21
Exception Parser

poulad:
This PR holds changes to exception parser API and feature goes along with changes in [Telegram.Bot.Extensions.Exceptions](https://github.com/TelegramBots/Telegram.Bot.Extensions.Exceptions) pacakge

### DIFF
--- a/src/Telegram.Bot/Exceptions/DefaultExceptionParser.cs
+++ b/src/Telegram.Bot/Exceptions/DefaultExceptionParser.cs
@@ -1,11 +1,15 @@
-﻿using System;
+using System;
 using System.Linq;
+using System.Net.Http;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using Telegram.Bot.Types;
 
 namespace Telegram.Bot.Exceptions
 {
-    internal static class ApiExceptionParser
+    internal class DefaultExceptionParser : IExceptionParser
     {
         private static readonly IApiExceptionInfo<ApiRequestException>[] ExceptionInfos = {
             new BadRequestExceptionInfo<ChatNotFoundException>("chat not found"),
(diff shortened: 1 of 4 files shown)


src/Telegram.Bot/Exceptions/DefaultExceptionParser.cs
```diff
@@ -93,5 +97,33 @@ private static string TryTruncateErrorDescription(string message, string descrip
...
+
+        /// <inheritdoc cref="IExceptionParser"/>
+        public async Task<Exception> ParseResponseAsync(HttpResponseMessage httpResponse)
+        {
+            Exception exception;
+
+            string responseJson = await httpResponse.Content.ReadAsStringAsync()
+                .ConfigureAwait(false);
+
+            var apiResponse = JsonConvert.DeserializeObject<JObject>(responseJson);
```
karb0f0s: Do I need to deserialize response in each custom Exception Handler? How is that better then processing already deserialized data in current implementation? At lest this gives an overhead of having dependency on `Newtonsoft.Json` in extension.

src/Telegram.Bot/Exceptions/DefaultExceptionParser.cs
```diff
@@ -93,5 +97,33 @@ private static string TryTruncateErrorDescription(string message, string descrip
...
+            else if (ok == false)
+            {
+                exception = new ApiRequestException(description);
+            }
+            else
+            {
+                exception = new ApiRequestException("Invalid status code: " + httpResponse.StatusCode);
+            }
+
+            return exception;
```
karb0f0s: Do I understand correctly, base parser will only return ApiRequestException? And it seems that `HttpStatusCode` was lost on the way :(

pouladpld:
@karb0f0s This is just PoC code and not real implementation. Feel free to change it.

tuscen:
Superseded by [Telegram.Bot.Extensions.Exceptions](TelegramBots/Telegram.Bot.Extensions.Exceptions) package.

PR #718, opened 2018-05-21
Properly disposing of HttpResponseMessages

MihaZupan:


pouladpld:
@MihaZupan Thank you for PR. I hold this for a while and will get back to it after we finalize changes in PR #717 .

PR #720, opened 2018-05-23
support for manually file stream reader

Ali-YousefiTelori:
I need to read file stream manually so I created this method.
I do fixes, created new method and test cases and change logs too.


MihaZupan:
Such a helper function has been discussed in the Telegram group. Doing this would mean not being able to properly dispose of the HttpResponseMessage object.

Ali-YousefiTelori:
Can you invite me on telegram group? i cannot join manually.
my username: @ali_visual_studio
my phone: 989390837919

PR #723, opened 2018-06-05
Fix message entity type serialization

tuscen:
This PR adds two new message entity type members: `PhoneNumber` and `Unknown`.
It also adds custom json converter for `MessageEntityType`. This ensures that deserialization of `MessageEntityType` is forward compatible and won't cause exception during deserialization.


karb0f0s:
#724

PR #731, opened 2018-06-07
Refactor chat info tests

poulad:


karb0f0s:
it is almost identical to #728 , but #728 covers more fixes to tests

karb0f0s:
@poulad  @tuscen  do we still need this?

tuscen:
Superseded by #917.

PR #734, opened 2018-06-07
Refactor MessageEntityType

tuscen:
This PR changes `MessageEntityType` to use `class` instead of `enum` to preserve original string value from the Bot API.

From a user standpoint not much is changed. But in case if Telegram again suddenly adds another entity type this change will preserve its original string value even through several json convertions so one could update its bot without waiting for the library update to see what the new type is.

```c#
// Old
var messageEntity = new MessageEntity
{
    Type = MessageEntityType.Hashtag,
    Offset = 10,
    Legnth = 10
};

switch (messageEntity.Type)
{
    case MessageEntityType.Hashtag:
    ...
}

// New
var messageEntity = new MessageEntity
{
    Type = MessageEntityType.Hashtag,
    Offset = 10,
    Legnth = 10
};

switch (messageEntity.Type.Type)
{
    case EntityType.Hashtag:
    ...
}
```

Naming could be tweaked a bit to make it look better. I don't like `.Type.Type` but couldn't come up with something better.


Banane9:
Maybe rename the last `.Type` to something with Enum?

tuscen:
I think this PR can be closed. This approach isn't substantially better compared to using strings and static classes for help.

PR #759, opened 2018-07-26
API 4.0: added EditMessageMedia method

MihaZupan:


src/Telegram.Bot/ITelegramBotClient.cs
```diff
@@ -875,15 +875,47 @@ public interface ITelegramBotClient
         /// <param name="parseMode">Send Markdown or HTML, if you want Telegram apps to show bold, italic, fixed-width text or inline URLs in the media caption.</param>
         /// <param name="replyMarkup">A JSON-serialized object for an inline keyboard.</param>
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
-        /// <returns><c>true</c> on success.</returns>
-        /// <see href="https://core.telegram.org/bots/api#editmessagecaption"/>
+        /// <returns>On success, the edited Description is returned.</returns>
```
tuscen: This is inline version of the method. It return true on success (Task in our case since there is no point in actually returning it).

src/Telegram.Bot/Requests/Update Messages/EditInlineMessageMediaRequest.cs
```diff
@@ -0,0 +1,44 @@
...
+using Telegram.Bot.Types.ReplyMarkups;
+
+// ReSharper disable once CheckNamespace
+namespace Telegram.Bot.Requests
+{
+    /// <summary>
+    /// Edit audio, document, photo, or video messages. On success the edited <see cref="Message"/> is returned.
+    /// </summary>
+    [JsonObject(MemberSerialization.OptIn, NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
+    public class EditInlineMessageMediaRequest : RequestBase<Message>,
```
tuscen: This is inline version of the method, it should be `RequestBase<bool>`.

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -962,6 +962,31 @@ await DownloadFileAsync(file.FilePath, destination, cancellationToken)
...
+            InlineKeyboardMarkup replyMarkup = default,
+            CancellationToken cancellationToken = default
+        ) =>
+            MakeRequestAsync(new EditMessageMediaRequest(chatId, messageId, media)
+            {
+                ReplyMarkup = replyMarkup
+            }, cancellationToken);
+
+        /// <inheritdoc />
+        public Task<Message> EditMessageMediaAsync(
```
tuscen: Same

src/Telegram.Bot/ITelegramBotClient.cs
```diff
@@ -875,15 +875,47 @@ public interface ITelegramBotClient
...
+        /// <summary>
+        /// Use this method to edit audio, document, photo, or video messages.
+        /// </summary>
+        /// <param name="inlineMessageId">Unique identifier of the sent message</param>
+        /// <param name="media">A JSON-serialized object for a new media content of the message</param>
+        /// <param name="replyMarkup">A JSON-serialized object for an inline keyboard.</param>
+        /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
+        /// <returns>On success the edited <see cref="Message"/> is returned.</returns>
+        /// <see href="https://core.telegram.org/bots/api#editmessagemedia"/>
+        Task<Message> EditMessageMediaAsync(
```
tuscen: Same, it should return `Task` without `Message` since it's inline version.

MihaZupan:
@tuscen Updated

src/Telegram.Bot/ITelegramBotClient.cs
karb0f0s: Can we also add overload method with correct argument order - `CancellationToken cancellationToken` as  the last parameter?

MihaZupan: Argument order was changed by @pouladpld in [this commit](https://github.com/TelegramBots/Telegram.Bot/commit/f58b5d5893200228ad65eb70b58cc6e14ee94959) to avoid a breaking change

MihaZupan: Speaking of which, is there any way to avoid doing the same with all the other newly added parameters?

karb0f0s: we can either introduce new update with breaking changes or add two implementations for each request - one with recommended parameter order and the other as it is now

pouladpld: @karb0f0s sounds good. we can use `[Obsolete]` to deprecate the overloaded methods with non-consistent param list.

pouladpld:
I merge this to bring the commits into our repo

PR #760, opened 2018-07-26
Cleanup

MihaZupan:
Cleanup branch for any changes I discover while updating to 4.0


src/Telegram.Bot/Types/Enums/MessageType.cs
```diff
@@ -72,6 +73,7 @@ public enum MessageType
         /// <summary>
         /// The <see cref="Message"/> contains a <see cref="Types.VideoNote"/>
         /// </summary>
+        [EnumMember(Value = "video_note")]
```
tuscen: By the way, this thing might be a breaking change. Someone might serialize update and/or message objects as is and store somewhere. And then rely on this values being in camelCase.

Issue #758, opened 2018-07-26
Bot API 4.0


karb0f0s:
# [Bot API 4.0.](https://core.telegram.org/bots/api#july-26-2018)

### Support for editing the media content of messages #764
- [X] method `editMessageMedia` #759 
- [X] `inputMediaAnimation` type #757 
- [X] `inputMediaAudio` type #757 
- [X] `inputMediaDocument` type #757 

### More new features

- [X] add field `thumb` to the `Audio` object #772
- [X] add support for attaching custom thumbnails to uploaded files #777
- [X] add the method `sendAnimation`, which can be used instead of `sendDocument` to send animations, specifying their duration, width and height. #777
- [X] add the field `animation` to the `Message` object. For backward compatibility, when this field is set, the document field will be also set. #768
- [X] new `MessageEntity` types: `cashtag` and `phone_number`. #761

### Support for Foursquare venues

- [X] new field `foursquare_type` to the objects `Venue`, `InlineQueryResultVenue` and `InputVenueMessageContent` #763 
- [X] new parameter `foursquare_type` to the `sendVenue` method. #763 

### vCard support when sharing contacts
- [X] new field `vcard` to the objects `Contact`, `InlineQueryResultContact`, `InputContactMessageContent` #762 
- [X] new `vcard` parameter for the method `sendContact`. #762

pouladpld:
First set of changes are released.
<https://github.com/TelegramBots/Telegram.Bot/pull/768>

karb0f0s:
#777 Telegram.Bot version 14.9.0 with Bot API 4.0 support released

PR #761, opened 2018-07-26
 API 4.0: Added a 'cashtag' entity type

MihaZupan:
Implement API 4.0: 'cashtag' entity type

#758


pouladpld:
What's a cashtag?

PR #764, opened 2018-07-27
Bot API 4.0: Edit Message Media method and new Input Media types

pouladpld:


pouladpld:
I believe there is an upstream bug that makes the test in `EditMessageMediaTests2` fail.

MihaZupan:
#765 should also be fine to merge

pouladpld:
@MihaZupan i'll include that in the next release soon.

PR #766, opened 2018-07-27
API 4.0: Telegram Passport - WIP

tuscen:
This PR adds Telegram Passport support.

#758


src/Telegram.Bot/Types/Enums/MessageType.cs
```diff
@@ -142,6 +143,12 @@ public enum MessageType
...
         /// The <see cref="Message"/> contains non-default <see cref="Message.MigrateToChatId"/>
         /// </summary>
-        MigratedFromGroup
+        MigratedFromGroup,
+
+        /// <summary>
+        /// The <see cref="Message"/> contains non-default <see cref="Message.PassportData"/>
+        /// </summary>
+        [EnumMember(Value = "passport_data")]
+        PassportData
```
pouladpld: We are trying not to use enum values anymore.

src/Telegram.Bot/Types/Message.cs
```diff
@@ -382,6 +389,9 @@ public MessageType Type
                 if (MigrateToChatId != default)
                     return MessageType.MigratedToSupergroup;
 
+                if (PassportData != default)
```
pouladpld: We are trying not to use enum values anymore.

pouladpld:
This PR won't be merged because scope is pretty big and changes min supported netstandard. I'll cherrypick some of the commits at #778

MihaZupan:
Some of the features here are only available in the higher target. Min target stays the same

src/Telegram.Bot/Enums/Gender.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// Gender, male or female
+        /// </summary>
+        public static class Gender
+        {
+            /// <summary>
+            /// Male
+            /// </summary>
+            public const string male = nameof(male);
```
tuscen: Are you sure about that? Public static fields should be in pascal case according to the MS [style guide](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-fields).

pouladpld: I was looking for readability but OK, I change it.

pouladpld:
#788

PR #771, opened 2018-07-29
More extensions

MihaZupan:
Extensions for interacting with files.
Overloads for methods that take both a `ChatId `and a `MessageId `=> Use `Message `instead


MihaZupan:
Closing in favour of #812, relevant overloads to add TBD.

PR #802, opened 2018-10-27
Add option to change the Token after initialization

MihaZupan:


poulad:
The bot client is immutable as of now. This feature makes it mutable.

MihaZupan:
See group chat history from today, proposed by @adelaiglesia

poulad:
Bot client is currently immutable. This change will make it mutable. Users might already have made assumptions on this.

I assumed the bot is the same in my bot framework for ASP.NET Core. However, a bot's token could be updated w/o restarting the web app process. [See configs docs]

Also, an overridden `OnConfiguring` method will also be called once similar to setting the token once from the ctor.

[See configs docs]: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.1

Antoniodelaiglesia:
#803

karb0f0s:
@MihaZupan can we close it?

PR #803, opened 2018-10-27
Add constructor decoupled configuration options

Antoniodelaiglesia:
With this PR configuration and construction is decoupled for better compatibility with IoC containers. We can continue using the TelegramBotClient as always, using its constructor:

`MyClient = new TelegramBotClient("1234567:myApiKey")`

Or now we can decouple configuration from constructor, allowing services to configure it from a dependency injection:

```
class MyService : IMyService {
    private readonly ITelegramBotClient _telegramBotClient;
    
    MyService(ITelegramBotClient telegramBotClient) {
        _telegramBotClient = telegramBotClient;

        _telegramBotClient.Configure(options =>
        {
            options.Token = "1234567:myApiKey";
        });
    }
}
```

This PR mantains the Bot client immutable, not allowing it to be configured twice. If you try to call TelegramBotClient.Configure after calling a parameterful constructor you will get a InvalidOperationException with this message: "**BotClient is already configured. This will happen if you used a constructor with parameters or if you called this method previously when using a parameterless constructor.**"

### DIFF
--- a/src/Telegram.Bot/Helpers/TelegramBotClientOptions.cs
+++ b/src/Telegram.Bot/Helpers/TelegramBotClientOptions.cs
@@ -1,0 +1,23 @@
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Text;
+
+namespace Telegram.Bot.Helpers
+{
+    /// <summary>
+    /// Used for configuring the bot 
+    /// </summary>
+    public class TelegramBotClientOptions
+    {
+        /// <summary>
+        /// The telegram bot token. Ask @botfather about it. 
+        /// </summary>
+        public string Token { get; set; } = "";
+        /// <summary>
+        /// An HttpBotClient instance that will be used for the bot. If not provided, a new one
+        /// will be created. 
+        /// </summary>
+        public HttpClient HttpClient { get; set; } = null;
+    }
+}
--- a/src/Telegram.Bot/TelegramBotClient.cs
+++ b/src/Telegram.Bot/TelegramBotClient.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Telegram.Bot.Args;
 using Telegram.Bot.Exceptions;
+using Telegram.Bot.Helpers;
 using Telegram.Bot.Requests;
 using Telegram.Bot.Requests.Abstractions;
 using Telegram.Bot.Types;
@@ -27,7 +28,6 @@ namespace Telegram.Bot
     public class TelegramBotClient : ITelegramBotClient
     {
         /// <inheritdoc/>
-        public int BotId { get; }
 
         private static readonly Update[] EmptyUpdates = { };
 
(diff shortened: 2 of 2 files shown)


poulad:
Using the approach shown above beats the purpose of the IoC.. How is that example use case different from this?

```c#
class MyService : IMyService {
    private readonly ITelegramBotClient _botClient = new TelegramBotClient("token");
} 
```

I don't know your exact use case but I guess what you really need is resolving different API tokens from the IoC container. The line of code for creating a bot client is always the same.

Even if you are trying to update a bot token easily, you can always do that using provider methods when registering the instance with IoC containers.

Alternatively, take a look at how I implemented the `BotBase` in the framework. It takes advantage of the `IOptions` configurations API:

<https://github.com/TelegramBots/Telegram.Bot.Framework/blob/0d0242904941e4ab9ef5348490060e04f5e67aa8/src/Telegram.Bot.Framework/BotBase.cs#L22-L23>

<https://github.com/TelegramBots/Telegram.Bot.Framework/blob/0d0242904941e4ab9ef5348490060e04f5e67aa8/sample/Quickstart.AspNetCore/EchoBot.cs#L6-L9>

Antoniodelaiglesia:
Hi, poulad! ☺

About your question:
> How is that example use case different from this?
```
class MyService : IMyService {
    private readonly ITelegramBotClient _botClient = new TelegramBotClient("token");
} 
```
It's very different because my example is not tighty coupled, which is yours because of the constructor. The IoC container shouldn't be the one which set the API key because it's not a configuration. Let me copy a message from the Telegram Group:

> The use case is that different services would want to use diferent api keys. If you check other .NET integrations with other services they leverage authentication / identification (which is what API Key is) to - frecuently - a method or a property so any service can configure it as it wish. It is an antipattern that the IoC container performs authentication / identification (API key) from a config file, because it is not a configuration. Bot behavior is the same despite of what API key you are using, so we can agree it is not a configuration. Each service should decide what identification it should provide, especially when you are writing a bot service (When a service must know the API key of an user of the system for using it)

poulad:
Again, what you are trying to solve is already implemented in our bot framework where you can have multiple bots(different tokens) within one ASP.NET Core app. It can also gives users the option to update the token of a bot.

However, for a quick fix, I suggest to add this to your IoC container:

```c#
class BotConfigsAccessor {
   string fooBotToken;
   string barBotToken;
}
```

and use it as in:

```c#
class MyService : IMyService {
    private readonly ITelegramBotClient _telegramBotClient;
    
    MyService(BotConfigsAccessor configs) { // or IOptions<T>
        _telegramBotClient = new TelegramBotClient(configs.fooBotToken);
    }
}
```

Implementing your example (different bot for authorized users) via the `IBot` from the bot framework, this could be done via resolving `FooBot` for regular users and `BarBot` for admins.

poulad:
@tuscen @MihaZupan @karb0f0s 
what do you think about this feature?

karb0f0s:
i don’t really like implementation  given in this pr

poulad:
> i don’t really like implementation given in this pr

We can talk about the implementation details later.
 
He presented a valid problem(using intances in IoC containers) but I'm trying to figure whether it makes sense for us to solve it in the project or leave it up to the users/framework.

poulad:
On that note, I'm closing this PR.

PR #814, opened 2019-04-13
add base_url and base_file_url to TelegramBotClient constructor parameter

naderAbolfazli:


src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -31,9 +31,9 @@ public class TelegramBotClient : ITelegramBotClient
 
         private static readonly Update[] EmptyUpdates = { };
 
-        private const string BaseUrl = "https://api.telegram.org/bot";
+        private string BaseUrl = "https://api.telegram.org/bot";
 
-        private const string BaseFileUrl = "https://api.telegram.org/file/bot";
+        private string BaseFileUrl = "https://api.telegram.org/file/bot";
```
Avend544: Man you have to change fields and mark them as readonly. You initizilithation BaseFileUrl and BaseUrl  in a constructor and when declare them.

MihaZupan:
Naming should follow the current style (no underscores in the middle of the name).
New parameters should also not come before the httpClient parameter, as it's less likely to ever be used and a breaking change for no reason.

poulad:
Hey @naderAbolfazli 
What's the reason for parametrizing base URLs?

naderAbolfazli:
> Hey @naderAbolfazli
> What's the reason for parametrizing base URLs?

hey
it's provided in some other languages telegram sdk such as python so those sdk could use for another api that likes telegram bot api. for example, in bale messenger new bot api released based on telegram bot api so bot developer could use their bot in both with only change of base_url

poulad:
In that case, we might want to set the base URL on the HTTP client and not introduce additional parameters to the constructor.

tuscen:
@poulad Wouldn't that require having two separate http clients since Bot API have a different url for files?

I don't mind having both base urls as properties with public setters.

MihaZupan:
> Wouldn't that require having two separate http clients since Bot API have a different url for files?

It could work for Telegram's api, setting the base as `"https://api.telegram.org/"`, but I don't think anything is gained since you still have to prepend `"bot/"` or `"file/bot/"` and other services may not share the same base for file urls.

+1 on public setters, anyone wanting to use this will surely be informed of having to change these values

naderAbolfazli:
public setters are ok?

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -31,9 +31,9 @@ public class TelegramBotClient : ITelegramBotClient
 
         private static readonly Update[] EmptyUpdates = { };
 
-        private const string BaseUrl = "https://api.telegram.org/bot";
+        private readonly string BaseUrl = "https://tapi.bale.ai/";
```
ndobryanskyy: seems like a malicious link

MihaZupan:
Closing as the issue has not been resolved for a while.

PR #822, opened 2019-06-01
Bot API 4.3

tuscen:


src/Telegram.Bot/Types/LoginUrl.cs
```diff
@@ -0,0 +1,36 @@
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Telegram.Bot.Types
+{
+    /// <summary>
+    /// This object represents a parameter of the inline keyboard button used to automatically authorize a user
+    /// </summary>
+    [JsonObject(MemberSerialization.OptIn, NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
+    public class LoginUrl
```
MihaZupan: Should we have a constructor accepting the Url to satisfy the required arguments here?
Other types seem to follow that rn.

tuscen: We actually don't have a consistent rule to have constructors for required params. Some types are follow this rule (e.g. `InlineQueryResult` and it's inheritors, all requests), but most types do not follow it.

Perhaps it would be a good thing to apply it consistently. Though I think we should just do it in v15 because it breaks everything.

PR #837, opened 2019-10-06
Add Telegram.Bot.Signed mirror project

MihaZupan:
I suppose this can be published along with the next package update.


src/Telegram.Bot.Signed/Telegram.Bot.Signed.csproj
```diff
@@ -0,0 +1,35 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
poulad: why is the version bumped to 2.0?

src/Telegram.Bot.Signed/Telegram.Bot.Signed.csproj
```diff
@@ -0,0 +1,35 @@
...
+    <VersionSuffix Condition="'$(VersionSuffix)' != ''">$(VersionSuffix)-$(APPVEYOR_BUILD_NUMBER)</VersionSuffix>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="..\Telegram.Bot\**\*.cs" Exclude="..\Telegram.Bot\obj\**;..\Telegram.Bot\bin\**">
+      <Link>%(RecursiveDir)%(Filename)%(Extension)</Link>
+    </Compile>
+  </ItemGroup>
```
poulad: can we sign the original assembly itself? i assume there are potential breaking changes.

MihaZupan: I was mostly following the pattern used by other projects (for example Markdig) to expose the signed version as Markdig.Signed.

I see the [posted guidance](https://docs.microsoft.com/en-gb/dotnet/standard/assembly/strong-named) is now NOT to ship two separate projects, though I think the argument listed there is weak.

As far as I know the only downside is that you shouldn't stop signing the assembly.

poulad:
Microsoft recommends having only 1 open-source library package and having it signed.

> ❌ DO NOT publish strong-named and non-strong-named versions of your library. For example, `Contoso.Api` and `Contoso.Api.StrongNamed`. [read more](https://docs.microsoft.com/en-gb/dotnet/standard/library-guidance/strong-naming).

Binding redirects can become annoying to our .NET Framework users 

> While strong naming a library has a number of benefits, it often results in runtime exceptions that an assembly can't be found and [requires binding redirects](https://docs.microsoft.com/en-gb/dotnet/framework/configure-apps/redirect-assembly-versions) in app.config/web.config to be fixed. [read more](https://docs.microsoft.com/en-gb/dotnet/standard/library-guidance/versioning#assembly-version).

karb0f0s:
@MihaZupan  can we close this due to Microsoft recommendations? https://github.com/TelegramBots/Telegram.Bot/pull/837#issuecomment-544737758

MihaZupan:
This specific PR yes, but we should consider some approach of signing a package so that users can consume it.

PR #840, opened 2019-10-15
Configure Azure Pipelines

poulad:
#### Added

- Azure Pipelines (see [sample build])
  - CI Artifacts
    ```
    dotnet add package Telegram.Bot --version 15.0.0-ci-232 --source "https://pkgs.dev.azure.com/poulad/_packaging/Telegram.Bot/nuget/v3/index.json"
    ```
  - Unit tests on all 3 major OSs with .NET Core 3.0 (see [test reports])
  - NuGet releases from `master` branch

#### Changed

- Updated Unit Tests project to .NET Core 3.0
- Edit deprecated element names in `csproj` file

#### Removed

- AppVeyor pipeline

[sample build]: https://dev.azure.com/poulad/Telegram.Bot/_build/results?buildId=232
[test reports]: https://dev.azure.com/poulad/Telegram.Bot/_build/results?buildId=232&view=ms.vss-test-web.build-test-results-tab


poulad:
I had some issues with running the integration tests but wasn't sure if it's the time to fix them.

There are a number of advanced xUnit capabilities that we have leveraged in our e2e tests. Those are not portable easily (stuck with netcore 2.0.0) and don't get much support or updates.

azure-pipelines.yml
```diff
@@ -0,0 +1,89 @@
...
+          - publish: .
+            artifact: windows-build
+          - bash: dotnet build --configuration Release --version-suffix "ci-${BUILD_BUILDID}+${BUILD_SOURCEVERSION}" src/Telegram.Bot/
+            displayName: "Build Package with CI version"
+          - task: NuGetCommand@2
+            displayName: "Publish CI Package to Azure Artifacts"
+            inputs:
+              command: push
+              publishVstsFeed: Telegram.Bot
+              packagesToPush: src/**/*-ci-*.nupkg
```
tuscen: Do we really need to publish nuget packages from each branch to the azure feed? And the order of packages will be completely meaningless.

poulad: Well, we can publish the artifacts for `develop` branch only and that provides sequential package versions.

Publishing a package for every single build enables the users to try out the latest features that are not stable, approved, or tested yet. We probably cannot guarantee the sequential order of the package versions but instead expect the users to be more involved in the build process and know what package version they are using.

tuscen: I'm not sure if this is needed outside of develop branch. Though I don't really mind it.

tuscen:
Substituted by #858.

PR #812, opened 2019-02-25
vNext draft

MihaZupan:
Major changes:
- [X] Target netstandard2.0 (allows us to remove `System.Net.Requests` as a dependency)
- [X] Remove polling events (replaced by extension package `Telegram.Bot.Extensions.Polling`)
- [X] Remove code marked as obsolete pre-v15
- [x] Remove exception parsing from core package
- [x] Fix parameter order (CT back to the end) / Use Request object

Minor:
- [X] Fix caption descriptions (#807)
- [x] Fix parameter types in `InputMediaAudio` (title, performer)
- [ ] File download overloads
- [x] Use doubles for coordinates?
- [x] Fix compilation warnings
- [x] Address ToDo comments
- [x] Review skipped test cases
- [x] Make test method styles look consistent
- [ ] Assembly signing
- [ ] Update the book and examples accordingly
- [x] Use C# 8 and enable nullable references

Under discussion:
- [ ] Use Request objects instead of parameter spaghetti

PR #843, opened 2019-11-05
Fixing the type of Title and Performer

MoishyFeder:
Title and Performer of an audio file are supposed to be string, not int.


MihaZupan:
Thank you, this change is already pending in the version 16 draft #812

PR #845, opened 2019-11-09
Make TelegramBotClient disposable

UselessToucan:
As far as I see, there is a `HttpClient` instance creation under the hood of a `TelegramBotClient`. Since `HttpClient` is `IDisposable`, I guess we should call `_httpClient.Dispose()` at some point.

### DIFF
--- a/src/Telegram.Bot/TelegramBotClient.cs
+++ b/src/Telegram.Bot/TelegramBotClient.cs
@@ -178,6 +178,7 @@ namespace Telegram.Bot
 
             _baseRequestUrl = $"{BaseUrl}{_token}/";
             _httpClient = httpClient ?? new HttpClient();
+            _isCustomHttpClientUsed = httpClient != null;
         }
 
         /// <summary>
@@ -212,6 +213,7 @@ namespace Telegram.Bot
         }
 
         private bool _isDisposed;
+        private readonly bool _isCustomHttpClientUsed;
 
         /// <summary>
         /// Allows a <see cref="TelegramBotClient"/> to try to free resources and perform other cleanup operations before it is reclaimed by garbage collection.
@@ -232,7 +234,8 @@ namespace Telegram.Bot
             if (_isDisposed)
                 return;
 
-            _httpClient.Dispose();
+            if (!_isCustomHttpClientUsed)
+                _httpClient.Dispose();
 
             _isDisposed = true;
         }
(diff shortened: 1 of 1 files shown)


tuscen:
This will make the client not DI-friendly since disposing of dependencies is a responsibility of a scope in a DI framework.

karb0f0s:
As mentioned in [various resources](https://stackoverflow.com/questions/15705092/do-httpclient-and-httpclienthandler-have-to-be-disposed-between-requests) - `HttpClient` should not be disposed after _each request_. But, it _should_ be disposed at the end of its owner's lifecycle.

Summarizing:
- we _should not_ dispose user-provided `HttpClient`
- we _might not_ care about constructor-instantiated `HttpClient` when `TelegramBotClient`  is a singleton object
- we _might_ consider disposing constructor-instantiated `HttpClient` when `TelegramBotClient` is a transient object 
- _maybe_ we should encourage users to instantiate `HttpClient` for us (either as singleton or via `HttpClientFactory`) leaving only constructor which accepts both `botToken` and `HttpClient`.

tuscen:
@karb0f0s Good points. It’s a complicated case with no single best strategy and I still don’t know what to do here.

On one hand forcing people to pass HttpClient manually every time will make using the client quite inconvenient. On the other, though, it will make it clearer who owns the client and who’s responsible for it’s disposal.

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -209,6 +212,36 @@ public TelegramBotClient(string token, IWebProxy webProxy)
...
+        /// <summary>
+        /// Disposes this <see cref="TelegramBotClient"/>
+        /// </summary>
+        public void Dispose()
+        {
+            dispose();
+            GC.SuppressFinalize(this);
+        }
+
+        private void dispose()
```
andy-dendy-repo, line 232: Rename method in conformity with naming conventions

wiz0u:
After studying the topic, I will solve this issue with PooledConnectionLifetime (in our .NET6+ build) which solves the issue with socket port exhaustion AND dns resolving whichever behavior the developer has with TelegramBotClient instance(s).
And no need to dispose it.

https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#recommended-use

Issue #36, opened 2015-09-05
Making requests when getting updates


9px:
hi. telegram [recently](https://core.telegram.org/bots/api#making-requests-when-getting-updates) add an ability to perform a request to the telegram Bot API while sending an answer to the webhook.
can you do a simple implement of this feature in your bot?

MrRoundRobin:
i'm working on this, but currently i'm not sure how i'll implement it.

a-a-k:
Hi,
could you show some samples of your work version to us?

mhverbakel:
The problem is that the called methods are all plain dictionaries at the moment. To be able to return them in MVC for example, you would need to switch to classes for method calls, and return an instance of that class using `this.Json(instance)` (in MVC).

Another option is making internal helper methods that generate the request parameters, and exposing them through public methods. Where you usually return with a `return SendWebRequestAsync`, you replace that by setting the `method` parameter and returning the dictionary. The client of the library can then handle outputting that in the webhook response.

mikhail-mikhaylov:
Maybe it's better to just return instance of some class that represents http response?
I mean some custom class, not HttpResponseMessage.
This class will have string body property, and some properies with headers. 
This would hide all implementation details and would not expose dictionary to the client.

Olfi01:
I have looked into this, and it seems that you've just got to return a json-encoded object containing all the parameters and the method name. So how about just adding a class for that? I'll try to do something like that.

tuscen:
Fixed by #847.

PR #847, opened 2019-11-21
Add support for requests in webhook responses

kipters:
This PR tries to solve #36 by adding two internal properties to RequestBase: the public boolean one controls the value returned by the internal one.
When serializing the request, if IsWebhookResponse is set to true, the internal property will have the same value as MethodName, so the serialized payload will include the correct `method` field. If it's set to false, `WebHookMethodName` will be equal to null and ignored by Json.Net.

It would be used like this:

```csharp
var request = new SendMessageRequest(chatId, text)
{
    IsWebhookResponse = true
};

return new OkObjectResult(request);
```

One could even create a special "OkWebhookRequestResult" that takes an IRequest<T> and automatically sets the property to true.

I've added two basic tests, is there anything else I would need to change/update?

### DIFF
--- a/src/Telegram.Bot/Requests/Abstractions/IRequest.cs
+++ b/src/Telegram.Bot/Requests/Abstractions/IRequest.cs
@@ -24,5 +24,10 @@ namespace Telegram.Bot.Requests.Abstractions
         /// </summary>
         /// <returns>Content of HTTP request</returns>
         HttpContent ToHttpContent();
+
+        /// <summary>
+        /// Allows this object to be used as a response in webhooks
+        /// </summary>
+        bool IsWebhookResponse { get; set; }
     }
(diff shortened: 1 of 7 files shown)


MihaZupan:
Nice! This looks like a really clean approach to me

tuscen:
Could you revert the last commit? Let's leave the new field only on `RequestBase` class.

kipters:
@tuscen Sure, I'm interested in the rationale behind this change though

tuscen:
@kipters It breaks compatibility for depending packages so we'd have to increment major version. I don't want to do it often so I want to postpone the interface change for later v16 release.

kipters:
@tuscen I understand, I've reverted the commit

PR #852, opened 2020-01-03
Implement Bot API 4.5

tuscen:


matteocontrini:
In file types they also changed the description for `file_id` from

>Identifier for this file

to

>Identifier for this file, which can be used to download or reuse the file

tuscen:
@matteocontrini Thanks. I've fixed descriptions for `file_id`.

PR #854, opened 2020-01-04
[WIP] 16.0.0

vova-lantsov-dev:
* **Removed events** `OnMessage`, `OnUpdate`, etc.
* **Removed methods** `StartReceiving()` / `StopReceiving()`
* **`HttpClient` is exposed publicly** for easier customization
* **All types, exceptions and interfaces moved** to `Telegram.Bot.Core` project
* **Support for multiple JSON serializers** (`System.Text.Json`, `Newtonsoft.Json`) based on `DataContractAttribute` and `DataMemberAttribute`
* **JsonConverter property** in `ITelegramBotClient`
* **Prepared for C# 8.0** (nullable reference types)
* **Prepared for `netcoreapp3.x`** (code analysis attributes such as `NotNullAttribute`)
* **Extensions for ASP.NET Core**

**TODO**
* Support for `JsonConstructorAttribute` for some classes
* Write new tests
* Complete the support for C# 8.0


src/Telegram.Bot.Core/Exceptions/ApiExceptionParser.cs
```diff
@@ -5,9 +5,10 @@
 
 namespace Telegram.Bot.Exceptions
 {
-    internal static class ApiExceptionParser
+    public static class ApiExceptionParser
```
tuscen: It should stay internal. We had an idea of making exception parsing customizable, but we aren't ready to commit to this API.

src/Telegram.Bot.Core/ITelegramBotClient.cs
```diff
@@ -24,80 +25,22 @@ public interface ITelegramBotClient
...
-
-        /// <summary>
-        /// Occurs before sending a request to API
-        /// </summary>
-        event EventHandler<ApiRequestEventArgs> MakingApiRequest;
-
-        /// <summary>
-        /// Occurs after receiving the response to an API request
-        /// </summary>
-        event EventHandler<ApiResponseEventArgs> ApiResponseReceived;
```
tuscen: These events should stay. They're meant for debugging purposes and for hooking up into request sending logic.

src/Telegram.Bot.Core/ITelegramBotClient.cs
```diff
@@ -119,19 +62,8 @@ public interface ITelegramBotClient
...
-        /// <exception cref="Exceptions.ApiRequestException"> Thrown if token is invalid</exception>
-        void StartReceiving(UpdateType[] allowedUpdates = null,
-                            CancellationToken cancellationToken = default);
-
-        /// <summary>
-        /// Stop update receiving
-        /// </summary>
-        void StopReceiving();
+        /// <inheritdoc cref="ITelegramBotJsonConverter"/>
+        ITelegramBotJsonConverter JsonConverter { get; set; }
```
tuscen: I don't think it should have public setter. JsonConverter can be accessed concurrently from more than one thread so there's a possibility of getting null reference exception if one thread attempts to overwrite it.

src/Telegram.Bot.Core/ITelegramBotJsonConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// </summary>
+    public interface ITelegramBotJsonConverter
+    {
+        /// <summary>
+        /// Deserializing json stream into Update class.
+        /// </summary>
+        /// <param name="jsonStream">UTF-8 stream which contains JSON data.</param>
+        /// <param name="ct">Cancellation token used for cancellation of the deserialization task.</param>
+        /// <returns>Update model.</returns>
+        ValueTask<Update> DeserializeAsync(Stream jsonStream, CancellationToken ct);
```
tuscen: Make it generic, don't bake in `Update` type here.

src/Telegram.Bot.Core/Requests/Abstractions/IRequest.cs
```diff
@@ -19,10 +21,16 @@ public interface IRequest<TResponse>
...
+        /// </summary>
+        ITelegramBotJsonConverter JsonConverter { get; set; }
+
         /// <summary>
         /// Generate content of HTTP message
         /// </summary>
+        /// <param name="ct"></param>
         /// <returns>Content of HTTP request</returns>
-        HttpContent ToHttpContent();
+        ValueTask<HttpContent> ToHttpContentAsync(CancellationToken ct);
```
tuscen: Same about token name

src/Telegram.Bot.Core/Requests/Available Methods/Admin Bot/DeleteChatPhotoRequest.cs
```diff
@@ -8,21 +7,20 @@ namespace Telegram.Bot.Requests
...
     {
         /// <summary>
-        /// Unique identifier for the target chat or username of the target channel (in the format @channelusername)
+        /// Initializes a new request with <see cref="ChatId"/> set to 0
         /// </summary>
-        [JsonProperty(Required = Required.Always)]
-        public ChatId ChatId { get; }
+        public DeleteChatPhotoRequest() : this(0)
+        {
+        }
```
tuscen: There shouldn't be a default ctor here or in any other request. They aren't meant to be serializable, they're library API and have required parameters in ctors without public setters.

src/Telegram.Bot.Core/Requests/Available Methods/Admin Bot/PromoteChatMemberRequest.cs
```diff
@@ -8,75 +7,66 @@ namespace Telegram.Bot.Requests
     /// <summary>
     /// Promote or demote a user in a supergroup or a channel. The bot must be an administrator in the chat for this to work and must have the appropriate admin rights. Pass False for all boolean parameters to demote a user.
     /// </summary>
-    [JsonObject(MemberSerialization.OptIn, NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
-    public class PromoteChatMemberRequest : RequestBase<bool>
+    public sealed class PromoteChatMemberRequest : ChatIdRequestBase<bool>
```
tuscen: Why is it sealed?

src/Telegram.Bot.Core/Exceptions/ApiExceptionParser.cs
vova-lantsov-dev: I made it public because it's placed in another assembly (`Telegram.Bot.Core`), but should be accessed from `Telegram.Bot`

So we have to decide what to do with this

src/Telegram.Bot.Core/ITelegramBotClient.cs
vova-lantsov-dev: Okay, gonna restore them

src/Telegram.Bot.Core/ITelegramBotClient.cs
vova-lantsov-dev: We can just notify everyone that this property is not thread safe. And we can implement something like ConcurrentTelegramBotClient which will be completely thread safe.

src/Telegram.Bot.Core/Requests/Available Methods/Admin Bot/DeleteChatPhotoRequest.cs
vova-lantsov-dev: Some guys may want to implement the bot's logic using json with reflection, like bot constructor. They need to have as comfortable as possible API.

src/Telegram.Bot.Core/Requests/Available Methods/Admin Bot/PromoteChatMemberRequest.cs
vova-lantsov-dev: Why not? It makes no sence. If someone needs to change the method name, he/she will implement a new class

src/Telegram.Bot.Core/Requests/ChatIdRequestBase.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using System.Net.Http;
+using Telegram.Bot.Types;
+
+namespace Telegram.Bot.Requests
+{
+    public abstract class ChatIdRequestBase<TResponse> : RequestBase<TResponse>
```
tuscen: Why is this class needed? We tried to minimise hierarchies and used interfaces for xml comments sharing instead.

src/Telegram.Bot.Core/Requests/RequestBase.cs
```diff
@@ -11,11 +14,9 @@ namespace Telegram.Bot.Requests
     /// <typeparam name="TResponse">Type of result expected in result</typeparam>
     public abstract class RequestBase<TResponse> : IRequest<TResponse>
     {
-        /// <inheritdoc />
         public HttpMethod Method { get; }
-
-        /// <inheritdoc />
-        public string MethodName { get; protected set; }
+        public string MethodName { get; }
+        public ITelegramBotJsonConverter JsonConverter { get; set; }
```
tuscen: `JsonConverter` is required for request serialization, so it has to be provided as a parameter in `ToHttpContentAsync`

vova-lantsov-dev: It's also used in methods of `FileRequestBase` class so I decided to implement it as property

src/Telegram.Bot.Core/Requests/ChatIdRequestBase.cs
vova-lantsov-dev: Is there a real performance reason? Because such class allows to write less code than when implementing the interface

src/Telegram.Bot.Core/Requests/Available Methods/Admin Bot/DeleteChatPhotoRequest.cs
tuscen: With reflection you can do basically everything. If someone uses reflection for that usecase they can add a bit more code to initialize types without default ctors and properties without setters.

src/Telegram.Bot.Core/Exceptions/ApiExceptionParser.cs
tuscen: `InternalVisibleToAttribute` on assembly will do the job

src/Telegram.Bot.Core/Requests/ChatIdRequestBase.cs
tuscen: It's not about performance, it's about minimizing public API space. No need to add stuff that don't bring much benefit.

If C# would have proper traits like Rust it would _probably_ make sense to separate shared properties into separate traits. In this case it's just a single class for a single property. Adding more such classes will have diminishing returns.

src/Telegram.Bot.Core/Requests/RequestBase.cs
tuscen: It should be a property on `IRequest.ToHttpContentAsync`. It's used literally everywhere and in effect is required.

tuscen:
My thoughts on this:

- The idea for abstracting away json serializer might work, but it relies havily on the assumption that `System.Text.Json` will support `DataContract` attributes in the future. We don't now yet if they would do it and it's a big if.

- Pulling in another third party json serializer is not the same as supporting `System.Text.Json`. It doesn't matter if it has the same API. It's still a third party solution with it's own assumptions on the future of `System.Text.Json`.

- As of now we can add support for `System.Text.Json` just by making a few breaking changes in optional properties and using conditional compilation to support it on `netstandard2.1+` and `netcoreapp3.0+`.

I can't commit to this API with `DataContracts` without knowing for sure that it will work with `System.Text.Json` in the future. I'll leave this PR open for future considerations.

TheMulti0:
Hi, is there any new progress with this important change?

vova-lantsov-dev:
@TheMulti0 no, you can complete it on your own by forking my repo. I'm doing another private implementation

PR #857, opened 2020-01-15
Custom API URLs support

igrmk:
This PR adds support for custom API URLs
It can be useful as a proxy to bypass local blocks for example

I have this nginx config on my proxy server that just redirects everything to api.telegram.org

    server {
        listen 443 ssl;
        ssl_certificate     <certificate_file>;
        ssl_certificate_key <certificate_key_file>;
        server_name api.telegram.example.com;
        location / {
            proxy_pass https://api.telegram.org;
        }
    }

So you can just configure a bot with **api.telegram.example.com** instead of **api.telegram.org** to bypass blocks

### DIFF
--- a/src/Telegram.Bot/TelegramBotClient.cs
+++ b/src/Telegram.Bot/TelegramBotClient.cs
@@ -31,9 +31,7 @@ namespace Telegram.Bot
 
         private static readonly Update[] EmptyUpdates = { };
 
-        private const string BaseUrl = "https://api.telegram.org/bot";
-
-        private const string BaseFileUrl = "https://api.telegram.org/file/bot";
+        private const string BaseUrl = "https://api.telegram.org";
 
         private readonly string _baseRequestUrl;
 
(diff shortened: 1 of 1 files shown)


MihaZupan:
This problem can be solved by providing a custom HttpClient

```c#

public static ITelegramBotClient CreateNewClient(string token)
{
    var httpClientHandler = new CustomHttpClientHandler();
    var httpClient = new HttpClient(httpClientHandler);
    return new TelegramBotClient(token, httpClient);
}

class CustomHttpClientHandler : HttpClientHandler
{
    protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
    {
        request.RequestUri = new Uri("https://different.service.domain" + request.RequestUri.AbsolutePath);
        return base.SendAsync(request, cancellationToken);
    }
}
```

This also offers a lot more flexibility, a user could use a different domain for file downloads, use different Url formats, etc.

Considering requests for this functionality aren't all that common, I don't think we should expose an API that makes assumptions about the custom Uri format.

PR #862, opened 2020-01-23
Implement Bot API 4.6

tuscen:
- [x] Add tests

### DIFF
--- a/src/Telegram.Bot/ITelegramBotClient.cs
+++ b/src/Telegram.Bot/ITelegramBotClient.cs
@@ -673,7 +673,7 @@ namespace Telegram.Bot
             IReplyMarkup replyMarkup = default,
             CancellationToken cancellationToken = default,
             bool? isAnonymous = default,
-            string type = default,
+            PollType? type = default,
             bool? allowsMultipleAnswers = default,
             int? correctOptionId = default,
             bool? isClosed = default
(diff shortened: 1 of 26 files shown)


src/Telegram.Bot/ITelegramBotClient.cs
```diff
@@ -656,16 +656,27 @@ public interface ITelegramBotClient
...
         /// <param name="replyMarkup">Additional interface options. A JSON-serialized object for an inline keyboard, custom reply keyboard, instructions to hide keyboard or to force a reply from the user.</param>
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
+        /// <param name="isAnonymous">True, if the poll needs to be anonymous, defaults to True</param>
+        /// <param name="type">Poll type, “quiz” or “regular”, defaults to “regular”</param>
+        /// <param name="allowsMultipleAnswers">True, if the poll allows multiple answers, ignored for polls in quiz mode, defaults to False</param>
+        /// <param name="correctOptionId">0-based identifier of the correct answer option, required for polls in quiz mode</param>
+        /// <param name="isClosed">Pass True, if the poll needs to be immediately closed</param>
         /// <returns>On success, the sent <see cref="Message"/> is returned.</returns>
         /// <see href="https://core.telegram.org/bots/api#sendpoll"/>
+        // TODO fix parameters order in vnext
```
karb0f0s: I'd suggest to mark this as an [Obsolete] and add overload with correct parameters order, so this one could be removed upon next release.

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -763,13 +763,23 @@ public Task<User> GetMeAsync(CancellationToken cancellationToken = default)
             bool disableNotification = default,
             int replyToMessageId = default,
             IReplyMarkup replyMarkup = default,
-            CancellationToken cancellationToken = default
```
karb0f0s: I'd suggest to mark this as an [Obsolete] and add overload with correct parameters order, so this one could be removed upon next release.

src/Telegram.Bot/ITelegramBotClient.cs
tuscen: Good idea 👍

tuscen: Unfortunately it requires both overloads to be different from the start, otherwise it leads to ambiguous invocation without passing optional arguments. There's only three required parameters and only one most natural order to them (at least for me).

PR #872, opened 2020-02-12
Add string presentation to Chat model

Dubzer:
I thought that It would be great if Chat model had string presentation, as already User model has it. This might be useful in logging


tuscen:
I don't think we'll accept this PR. We might remove `ToString` override from `User` object in vnext.

PR #897, opened 2020-04-28
- ChatId constructor fix

Backs:
More readable user name detection and null check


tuscen:
@karb0f0s good points.

Regarding the PR, @Backs, you changed the contract by removing the check for username length. If it is needed or not is a separate question that probably should be addressed.  Also throwing `ArgumentNullException` also changes the contract. If we to make these changes it would be a breaking change regarding runtime behavior and should be accompanied with major version bump.

I'd leave this PR open until I continue to v16.

Backs:
Added validation accoring to https://telegram.org/faq and tests

Backs:
Something else?

tuscen:
@Backs Why did you remove parsing of numerical ids? Could you bring them back, but leave last `else` clause with `ArgumentException`? Everything else seems fine.

As I said, I can't merge any changes that affect runtime behaviour right now. I'll merge this PR when I continue with v16.

As an idea I think regex might be a better way to check for valid usernames. We can stick to the way client API docs constrains them:

> Accepted characters: A-z (case-insensitive), 0-9 and underscores.
Length: 5-32 characters.

This regex should work: `^@[a-zA-Z0-9_]{5,32}`. You can preallocate a private static field with this regex (since regexes are immutable and thread safe) and use it to match for valid usernames in the ctor.

Backs:
@tuscen regex is ok. But ChatId has constructors for int and long. Does it really need to parse int and long values from string?

tuscen:
@Backs It's just convenient to have this. Bot API also works this way, you can pass numerical chad id in json as a string and it'll work.

Backs:
Fixed. Regex can not be compiled due to netstandart 1.1

tuscen:
@Backs Could you reapply your PR to v15 branch? Most work on the next major version happens there. You can just close this PR and make a new one with a single commit with all the changes.

PR #913, opened 2020-07-04
Refactor how errors are handled

tuscen:
This PR refactors how request sending logic handles errors.

- A new type of base exception is introduced - `RequestException`. It's thrown in situtations when there is no correct error from Bot API. E.g. on serialization errors, DNS errors, errors from badly configured proxies, etc. If possible `HttpStatusCode` and stringified body of the response are included in the exception. In practice `ITelegramBotClient.MakeRequestAsync` should not throw any other exception than `RequestException` and it's inheritors.
- `ApiRequestException` extends `RequestException`.
- `IExceptionParser` interface is introduced to offload exceptions parsing into a separate package. This way breaking change of removing all custom exceptions can be mitigated by installing an additional package where current custom exceptions will be put.
- Exceptions from user code are not wrapped inside `RequestException` since these exceptions are not related to request sending mechanism. User code is event handlers inside `MakingApiRequest` and `ApiResponseReceived` events and custom implementations of `IExceptionParser`.
- Two new internal types are added for handling `200` and non-`200` responses from Bot API. These types are variations of `ApiResponse<T>` but with different required properties. Depending on the status code of the response serialization exception will be thrown, handled and and rethrown wrapped inside `RequestException`.
- A new method for making requests is introduced - `ITelegramBotClient.SendRequestAsync<ApiResponse<TResult>>`. This method is exceptionless. That is correct errors from Bot API are not thrown as `ApiRequestException`, but are returned as `ApiResponse<TResult>` instance. If there's no response from Bot API that can be deserialized correctly or there is other problems `RequestException` will be thrown.


src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -104,54 +113,130 @@ public TelegramBotClient(string token, HttpClient? httpClient = null)
...
-                default:
-                    httpResponse.EnsureSuccessStatusCode();
-                    break;
+            var httpRequest = new HttpRequestMessage(request.Method, url)
+            {
+                Content = request.ToHttpContent()
+            };
+
+            var reqDataArgs = new ApiRequestEventArgs(request.MethodName, httpRequest.Content);
+            MakingApiRequest?.Invoke(this, reqDataArgs);
```
MihaZupan, line 198:
```suggestion
            MakingApiRequest?.Invoke(this, new ApiRequestEventArgs(request.MethodName, httpRequest.Content));
```

Avoid allocating EventArgs if there are no subscribers (should be the same in `MakeRequestAsync` above

src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -104,54 +113,130 @@ public TelegramBotClient(string token, HttpClient? httpClient = null)
...
+                        failedApiResponse.ErrorCode,
+                        failedApiResponse.Description,
+                        failedApiResponse.Parameters
+                    );
+                }
+
+                var successfulApiResponse = await httpResponse
+                    .DeserializeContentAsync<SuccessfulApiResponse<TResult>>(
+                        actualResponseStatusCode
+                    );
```
MihaZupan, line 163: Shouldn't ever take the async path here anyway, but `.ConfigureAwait(false);` for consistency

src/Telegram.Bot/TelegramBotClient.cs
tuscen: I've look into it and there might be a downside with that particular solution. With the current setup one could actually track requests because the same instance of `reqDataArgs` then passed inside `ApiResponseEventArgs`. So one could just store event args from `MakingApiRequest` in a dictionary (e.g. with additional metadata like req start time) and on `ApiResponseReceived` associated metadata because of reference equality. I don't know if that was Poulad's original idea, but that's how it's ended up.

poulad: That's right @tuscen. The intent was to have a way for users to correlate these 2 request and response events. Object reference is kinda a correlation ID between these 2 separate events.

PR #914, opened 2020-07-08
Remove defensive copying of status code

tuscen:
Also in PR:
- Ensure the response is disposed in `DownloadFileAsync` in case of exception
- Handle `TaskCancelledException` in `SendRequestAsync`

### DIFF
--- a/src/Telegram.Bot/TelegramBotClient.cs
+++ b/src/Telegram.Bot/TelegramBotClient.cs
@@ -143,10 +143,6 @@ namespace Telegram.Bot
 
             try
             {
-                // required since user might be able to set new status code using following
-                // event arg
-                var actualResponseStatusCode = httpResponse.StatusCode;
-
                 if (ApiResponseReceived != null)
                 {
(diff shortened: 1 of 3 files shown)


src/Telegram.Bot/TelegramBotClient.cs
```diff
@@ -103,8 +103,7 @@ public TelegramBotClient(string token, HttpClient? httpClient = null)
             IRequest<TResult> request,
             CancellationToken cancellationToken = default)
         {
-            var url = _baseRequestUrl + request.MethodName;
-
+            var url = new Uri($"{_baseRequestUrl}{request.MethodName}", UriKind.Absolute);
```
karb0f0s, line 106: what about avoiding interpolation by storing _baseRequestUrl as Uri and use `Uri(Uri,String) constructor instead?

```csharp
var url = Uri(_baseRequestUrl, request.MethodName);
```

https://docs.microsoft.com/en-us/dotnet/api/system.uri.-ctor?view=netcore-3.1#System_Uri__ctor_System_Uri_System_String_

PR #931, opened 2020-10-03
Add unit test for TelegramBotClient.StartReceiving

nminaya:
Add test for when StartReceiving method is called with a cancelled Token passed as an argument.


nminaya:
Closed because wrong branch to merge into

PR #932, opened 2020-10-03
Add unit test for TelegramBotClient.StartReceiving

nminaya:
Add test for when StartReceiving method is called with a cancelled Token passed as an argument.

### DIFF
--- a/test/Telegram.Bot.Tests.Unit/Framework/RequestCounterHttpClientHandler.cs
+++ b/test/Telegram.Bot.Tests.Unit/Framework/RequestCounterHttpClientHandler.cs
@@ -1,0 +1,27 @@
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Telegram.Bot.Tests.Unit.Framework
+{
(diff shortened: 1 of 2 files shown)


tuscen:
Thanks for the contribution. There's some problems in your test. 

1. It doesn't really test anything since you cancel the token before polling starts
2. It polls real bot api servers, which it shouldn't. What you need to do is to write a mock for `HttpMessageHandler` that manages http requests and returns dummy data with update in a body. You then need to pass it into `HttpClient` and pass that `HttpClient` into `TelegramBotClient`.
3. It would also be good to ensure that the client makes a request for updates at least once, so there might be a counter or something or how many requests were done and a slight delay between polling start and stop to ensure that at least some work was done. At the end of the test you'll need to assert the the number of request is greater than 0.

nminaya:
@tuscen Done. But the point of that test is making sure we don't make ANY request since we received a cancelled Token... So I added a mocked HttpClient to make sure that, and another test to make sure the opposite.

karb0f0s:
with a release of v17 we're removing events. these tests are not relevant any more.

PR #933, opened 2020-10-06
Added ChannelPost event

V1rusFalcon:
Hi, two days ago i figured out that was missing the channelpost event and all the classes about it, i tried to implement it. Probably is missing something.


tuscen:
There's no need to add a separate type for channel messages. The PR is also breaks backwards compatibility and targets incorrect branch. I'll close it.

PR #945, opened 2020-12-27
Telegram.Bot.CommandHandler and Telegram.Bot.Interactivity Integrations

AliBdeir:
This pull request adds two libraries, Telegram.Bot.CommandHandler and Telegram.Bot.Interactivity to the repository.

**Telegram.Bot.CommandHandler** allows for the creation of command modules, making listening for and responding to commands as simple as:
```cs
public class Commands : CommandModule
{
    [Command("ping")]
    public async Task Ping(CommandContext ctx)
    {
        await ctx.RespondAsync("Pong!");
    }
}
```

**Telegram.Bot.Interactivity** allows for bot-user comandless interactions, making it easy to wait for a response from a user:
```cs
private static async void OnMessageSent(object sender, MessageEventArgs e)
{
    if (e.Message.Text == "/hello")
    {
        //Ask the user for his name
        await botClient.SendTextMessageAsync(e.Message.Chat.Id, "Hello! What's your name?");
        // Get the client's interactivity object
        var interactivity = botClient.GetInteractivity();
        //Wait for a message.
        var result = await interactivity.WaitForMessageAsync(e.Message.Chat, e.Message.From);
        if (result.Value != null)
        {
            // The user responded in time.
            await botClient.SendTextMessageAsync(e.Message.Chat.Id, $"Hello, {result.Value}!");
        }
        else if (result.TimedOut)
        {
            // The user didn't respond within the set DefaultTimeOutTime, which is 2 minutes by default unless changed in the Configuration.
            await botClient.SendTextMessageAsync(e.Message.Chat.Id, $"The process timed out.");
        }
        else
        {
            // An error has occured.
            await botClient.SendTextMessageAsync(e.Message.Chat.Id, $"An unknown error has occured.");
        }
    }
}
```


Fedorus:
This library is considered as wrapper to API so I doubt that this will be merged. You can keep this as separate libraries and add example of its usage to [Example ](https://github.com/TelegramBots/Telegram.Bot.Examples) repo

TheMulti0:
Excellent addition!
I'd love to see this as an external library, since I don't see the folks here adding functionality like this that isn't core behavior.

tuscen:
I don't think this should be a part of the core client. It's perfectly fine for it to be in an external library.

PR #970, opened 2021-04-01
Add constructor for Local Bot API Server

zefir01:
This is need to use Local Bot API Server (https://core.telegram.org/bots/api#using-a-local-bot-api-server) or for use a reverse proxy.


Fedorus:
1. It is already implemented in 5.0+ version
2. It is not working correctly with files.

zefir01:
It is not working correctly with files.
Not working correctly with files in 5.0+ version? or in my request?

Fedorus:
In your PR. 
Cuz with --local you will get local path to file on your PC instead of https link to telegram servers and that will break few method in the library

zefir01:
ok. i got it. sorry.

PR #975, opened 2021-04-17
Mark polling methods and events as obsolete

karb0f0s:


karb0f0s:
should [event args classes](https://github.com/TelegramBots/Telegram.Bot/tree/api5_1/src/Telegram.Bot/Args) also be marked as obsolete?

tuscen:
@karb0f0s I'd say they should be since they are a part of events.

karb0f0s:
> @karb0f0s I'd say they should be since they are a part of events.

done

PR #985, opened 2021-05-02
Bot API 5.2 (April 26, 2021)

karb0f0s:
#980

### DIFF
--- a/src/Telegram.Bot/Types/VoiceChatScheduled.cs
+++ b/src/Telegram.Bot/Types/VoiceChatScheduled.cs
@@ -1,0 +1,19 @@
+using System;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace Telegram.Bot.Types
+{
+    /// <summary>
+    /// This object represents a service message about a voice chat scheduled in the chat.
+    /// </summary>
+    public class VoiceChatScheduled
+    {
+        /// <summary>
+        /// Point in time (Unix timestamp) when the voice chat is supposed to be started by a chat administrator
+        /// </summary>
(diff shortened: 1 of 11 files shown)


src/Telegram.Bot/Types/VoiceChatScheduled.cs
```diff
@@ -10,10 +10,9 @@ namespace Telegram.Bot.Types
     public class VoiceChatScheduled
     {
         /// <summary>
-        /// Point in time (Unix timestamp) when the voice chat is supposed to be started by a chat administrator
+        /// Point in time when the voice chat is supposed to be started by a chat administrator
         /// </summary>
-        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
```
tuscen: Why did you remove JsonProperty attribute?

karb0f0s: as I understand it `start_date ` is not optional in [VoiceChatScheduled](https://core.telegram.org/bots/api#voicechatscheduled), but `VoiceChatScheduled` might be optional inside [`Message`](https://core.telegram.org/bots/api#message), so it's marked as optional [there](https://github.com/TelegramBots/Telegram.Bot/blob/925532d241463ade1730e35de4a3220b7545f27b/src/Telegram.Bot/Types/Message.cs#L349-L350)

tuscen: It also lack `[JsonObject(MemberSerialization.OptIn, NamingStrategyType = typeof(SnakeCaseNamingStrategy))]` attribute. We use opt-in strategy everywhere os `JsonProperty` attributes must be applied.

tuscen (approved):

PR #998, opened 2021-07-03
Some Boolean return types was missed

immmdreza:
Some methods should return `Task<bool>` but they do `Task`
- `PinChatMessageAsync`
- `UnpinChatMessageAsync`
- `UnpinAllChatMessages`
- `SetChatStickerSetAsync`
- `DeleteChatStickerSetAsync`


tuscen:
It is intentional since these `true` value doesn't mean anything. If the request is successful nothing happens in the runtime, otherwise there will be exception.

immmdreza:
> It is intentional since `true` values is these requests doesn't mean anything. If the request is successful nothing happens in the runtime, otherwise there will be exception.

That’s ok then.

PR #1075, opened 2022-03-24
update_id need change type to bigger capacity

Ant1gon:
update_id is better to make a long type. because in large chatbots the capacity of the int type ends quickly.


Fedorus:
This field relies on telegram documentation.
https://core.telegram.org/bots/api#update

As they state its not always sequential, so there shouldn't be overflow.

Ant1gon:
but. it's happened =( and its had happen no one time
snap_screen_20220324170320

karb0f0s:
we cannot just make something *better*. we follow the specification, which states that update_id is integer value. unfortunately, your suggestion only breaks compatibility without giving any real value.

Ant1gon:
ok. youre right

PR #1078, opened 2022-04-09
[WIP, lib v19] Source generators for Types and Requests with System.Text.Json support

vova-lantsov-dev:
## Summary
This PR introduces 2 source generators that generate the classes for Types and Requests folders. Also it brings an API parser app.

### API parser
It produces a JSON file according to the following schema: https://github.com/TelegramBots/Telegram.Bot/blob/abdab6f0fdc5dcf9e84364dfc431abce766ddcf4/src/Telegram.Bot.ApiParser/schema.json

### Types source generator
Scriban template that will be rendered for each Bot API type: https://github.com/TelegramBots/Telegram.Bot/blob/abdab6f0fdc5dcf9e84364dfc431abce766ddcf4/src/Telegram.Bot.Generators/Templates.Types.cs#L16-L69

where:
- `{{ type_namespace }}` is namespace (starts with `Telegram.Bot.Types` but can additionally contain sub-namespace such as `Telegram.Bot.Types.InlineQueryResults`)
- `{{ type_description }}` is type's multi-line description parsed from official Bot API page
- `{{ type_name }}` is the `class`'es name, for example for Bot API type `Chat` we will generate `Chat` class located in file `Chat.generated.cs`
- `{{ parameter.parameter_description }}` is property (parameter) multi-line description parsed from official Bot API page
- `{{ parameter.parameter_name }}` is snake_case name of parameter parsed from official Bot API page
- `{{ parameter_type_name }}` is fully qualified type name of property, including its namespace (if needed)
- `{{ pascal_case_name }}` is property name in PascalCase

The main concepts of this generation are:
- `JsonIgnore` attribute as well as Newtonsoft's Required and DefaultValueHandling arguments are only added when necessary
- Generated classes are partial so they can be extended with custom logic
- If class with corresponding name already exist and IS NOT partial - we skip generation of such class. If it exists and IS partial - class will be generated.

At the moment, the source generator looks for `class` types with `public` and without `static` modifiers, located inside `Types` folder and all sub-folders.

#### TODO
- [x] Add support for Required and Ignore attributes/arguments
- [ ] Extend new Enum source generator with STJ serializers
- [ ] Allow sub-namespaces for some types (such as `Telegram.Bot.Types.InlineQueryResults`)
- [x] Add support for custom mapping of type names (f.e. `ForceReply` -> `ForceReplyMarkup`)
- [ ] Write unit tests

### Requests source generator
TODO


vova-lantsov-dev:
@tuscen @karb0f0s @MihaZupan let's discuss this feature if we can bring it in `18.0.0` release

Lorymi:
Any update?

tuscen:
Unfortunately, I don’t have enough free time at the moment to review and test this PR.

ductai230894:
wait update to change System.Text.Json

PR #1087, opened 2022-04-28
Add options class to configure the client

tuscen:
Also adds ability to use test environment


Fedorus:
I see a lot of new 
`using JetBrains.Annotations;`
do we really need that?

tuscen:
> I see a lot of new `using JetBrains.Annotations;` do we really need that?

They are compile time only. It's so Rider won't suggest refactoring to remove types because it thinks they are unused.

karb0f0s (approved):

PR #1093, opened 2022-04-30
Enum converter source generator

karb0f0s:
It's a prototype for Enum Converter source generator.

It works, but has a few caveats:

1. It assumes that zero (0) element is always Unknown.
2. This leads to conflicts with `MessageType.Unknown`, `UpdateType.Unknown`, `InlineQueryResultType.Unknown`.
3. It assumes that serialized value has snake case format. It doesn't work for special cases: `ParseMode` is not snake case and `Emoji` has "graphic" representation.

Generator definitely requires tests for all those cases.

Possible ways to work around these issues 
- Add attribute parameter or special attribute `[NoUnknown]` to avoid generation of Unknown value.
- Check for `Unknown` value in source generator and skip adding default value.
- Explicitly add `Unknown` to all enums
- Add custom attribute (for example `[Display(":banana:")]` for special cases.


karb0f0s:
The generator is based on Andrew Lock's [NetEscapades.EnumGenerators](https://github.com/andrewlock/NetEscapades.EnumGenerators)

karb0f0s:
I fixed a bug for `Unknown` enum field. Still, our API is inconsistent - some enums have Unknown field, some not. Generator will always emit "unknown" and 0 for bad values.
Also, I added a generator that emits a converter based on switch expression and run a benchmark for it:
https://github.com/TelegramBots/Telegram.Bot/blob/24d2970cec6991c4af7616b6b322e231d46b1ddb/src/EnumSerializer.Generator/SourceGenerationHelper.cs#L117

`GetStringValue` is three times faster and `GetEnumValue` just a tiny bit faster:

|                    Method |      Mean |     Error |    StdDev | Allocated |
|-------------------------- |----------:|----------:|----------:|----------:|
|   GetStringValueDictionary |  9.633 ns | 0.0849 ns | 0.0794 ns |         - |
|       GetStringValueSwitch |  3.055 ns | 0.0307 ns | 0.0287 ns |         - |
| GetEnumValueDictionary | 16.786 ns | 0.0708 ns | 0.0662 ns |         - |
|     GetEnumValueSwitch | 14.051 ns | 0.0728 ns | 0.0681 ns |         - |

But I was not satisfied with StringBuilder approach - too much bloat, too hard to get adequate result, so I added [@x](https://github.com/xoofx)'s [scriban](https://github.com/scriban/scriban) template engine and the result is this beautiful method:

https://github.com/TelegramBots/Telegram.Bot/blob/24d2970cec6991c4af7616b6b322e231d46b1ddb/src/EnumSerializer.Generator/SourceGenerationHelper.cs#L165

tuscen (approved):

PR #1095, opened 2022-05-02
Enhance nullability for ChatId conversion

zvirja:
Enhance nullability for `ChatId` implicit cast operator, so that compiler does not complain when `Chat` is not null.


tuscen (approved):

tuscen:
Thanks!

Issue #1109, opened 2022-06-21
Bot API 6.1 (June 20, 2022)


karb0f0s:
# [Bot API 6.1](https://core.telegram.org/bots/api#june-20-2022)

## Media in Descriptions

- Added support for photos and videos in the 'What can this bot do?' section (shown on the bot's start screen). Use [BotFather](https://t.me/BotFather) to set up media.

## Web App Improvements

- Added the fields `version`, `headerColor`, `backgroundColor`, `BackButton`, `HapticFeedback` and the methods `isVersionAtLeast`, `setHeaderColor`, `setBackgroundColor`, `openLink`, `openTelegramLink`, `openInvoice` to the class `WebApp`.
- Added the field `secondary_bg_color` to the class `ThemeParams`.
- Added the method `offClick` to the class `MainButton`.
- Added the fields `chat`, `can_send_after` to the class `WebAppInitData`.
- Added the events `backButtonClicked`, `settingsButtonClicked`, `invoiceClosed`.

## Join Requests & Payments

- Added the fields `join_to_send_messages` and `join_by_request` to the class `Chat`.
- Added the ability to process join requests which were created without an invite link. Bots will receive a “`chat_join_request`” update as usual.
- Added the method `createInvoiceLink` to generate an HTTP link for an invoice.

## Telegram Premium Support (more info)

- The maximum value of the field `file_size` in the classes `Animation`, `Audio`, `Document`, `Video`, `Voice`, and `File` can no longer be stored in a signed 32-bit integer type. This change is necessary to support 4GB files uploaded by premium accounts.
- Added the field `is_premium` to the class `User`.
- Added the field `premium_animation` to the class `Sticker`.

## Attachment Menu Integration

- Added the field `added_to_attachment_menu` to the class `User`.
- Bots integrated in the attachment menu can now be used in groups, supergroups and channels.
- Added support for t.me links that can be used to select the chat in which the attachment menu with the bot will be opened.

## Other Changes

- Added the parameter `secret_token` to the method `setWebhook`.
- As previously announced, only HTTPS links are now allowed in `login_url` inline keyboard buttons.

Umidrifkatov:
how to use HapticFeedback?

Umidrifkatov:
> how to use HapticFeedback?

Just figured out it. 
`window.Telegram.WebApp.impactOccurred("medium")`

Artem990:
Which link I have to pass to - openInvoice(url[, callback])

[...]

PR #1113, opened 2022-06-24
BotApi 6.1

karb0f0s:
Implement BotApi 6.1 
#1109


src/Telegram.Bot/TelegramBotClientExtensions.ApiMethods.cs
```diff
@@ -152,6 +158,7 @@ await botClient.ThrowIfNull(nameof(botClient))
         int? maxConnections = default,
         IEnumerable<UpdateType>? allowedUpdates = default,
         bool? dropPendingUpdates = default,
+        string? secretToken = default,
```
tuscen, line 161: Oops, that is a breaking change. It should be v19 then.

tuscen (approved):

PR #1115, opened 2022-07-05
Added native_ui property support to AnswerCallbackQuery

Nypifok:
https://core.telegram.org/api/bots/buttons#callback-queries
That property allows to specify whether to open game URLs in a WebView or in native UI.


karb0f0s:
this option is not exposed to the [Bot API](https://core.telegram.org/bots/api#answercallbackquery) and we cannot implement it.

PR #1121, opened 2022-08-06
align InputFile type hierarchy with Bot API

karb0f0s:


tuscen:
I think it's safe to use `record struct` with primary ctos for all new `InputFile`s since they don't emit `init` that reqruires `IsExternalUnitAttribute`.

tuscen (approved):

PR #1146, opened 2022-12-02
Add dependabot

karb0f0s:
Track dependency changes with dependabot action


tuscen:
I think we can also add cargo tracking to update mdbook

tuscen (approved):

Issue #1156, opened 2022-12-06
Make InputFileUrl ctor accept Uri argument too (like  InputOnlineFile did)


karb0f0s:
Make `InputFileUrl` ctor accept Uri argument too (like  `InputOnlineFile` did)

PR #1158, opened 2022-12-07
Make InputFileUrl ctor accept Uri

karb0f0s:
#1156

### DIFF
--- a/src/Telegram.Bot/Types/InputFiles/InputFileUrl.cs
+++ b/src/Telegram.Bot/Types/InputFiles/InputFileUrl.cs
@@ -21,4 +21,11 @@ public class InputFileUrl : IInputFile
     /// </summary>
     /// <param name="url"></param>
     public InputFileUrl(string url) => Url = url;
+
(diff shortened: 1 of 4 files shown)


src/Telegram.Bot/Types/InputFiles/InputFileUrl.cs
```diff
@@ -4,21 +4,28 @@
...
 public class InputFileUrl : IInputFile
 {
     /// <inheritdoc/>
     public FileType FileType => FileType.Url;
 
     /// <summary>
-    ///
+    /// HTTP URL for the file to be sent
     /// </summary>
     public string Url { get; }
```
tuscen, line 17: I think we should just make it a Uri instance and accept it everywhere we accepted strings as URIs

karb0f0s: we don't have any `Uri` typed properties in any other classes though

tuscen (approved):

PR #1172, opened 2022-12-16
Fix culture warnings

karb0f0s:
Depends on #1171 

Improve string comparison and conversion methods forcing `CultureInfo.InvariantCulture`

### DIFF
--- a/src/Telegram.Bot/Types/ChatId.cs
+++ b/src/Telegram.Bot/Types/ChatId.cs
@@ -40,17 +40,21 @@ public class ChatId : IEquatable<ChatId>
     public ChatId(string username)
     {
         if (username is null) { throw new ArgumentNullException(nameof(username)); }
-        if (username.Length > 1 && username.StartsWith("@", StringComparison.InvariantCulture))
+        if (username.Length > 1 && username[0] == '@')
         {
             Username = username;
         }
-        else if (long.TryParse(username, out var identifier))
+        else if (long.TryParse(
+            s: username,
+            style: NumberStyles.Integer,
+            provider: CultureInfo.InvariantCulture,
+            result: out var identifier))
         {
             Identifier = identifier;
         }
         else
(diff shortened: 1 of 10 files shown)


src/Telegram.Bot/Types/ChatId.cs
```diff
@@ -40,17 +40,21 @@ public class ChatId : IEquatable<ChatId>
     public ChatId(string username)
     {
         if (username is null) { throw new ArgumentNullException(nameof(username)); }
-        if (username.Length > 1 && username.StartsWith("@", StringComparison.InvariantCulture))
+        if (username.Length > 1 && username[0] == '@')
```
tuscen, line 43: Since we now use C# 11 we have access to list patterns.

```suggestion
        if (username.Length > 1 && username[0] == '@')
        if (username is ['@', ..])
```

karb0f0s: This change is only possible in net6.0 target. Should I add a directive for it?

Severity | Code | Description | Project | File | Line | Suppression State
-- | -- | -- | -- | -- | -- | --
Error (active) | CS0518 | Predefined type 'System.Index' is not defined or imported | Telegram.Bot (netstandard2.0) | Telegram.Bot\src\Telegram.Bot\Types\ChatId.cs | 47 |  
Error (active) | CS0656 | Missing compiler required member 'System.Index.GetOffset' | Telegram.Bot (netstandard2.0) | Telegram.Bot\src\Telegram.Bot\Types\ChatId.cs | 47 |

karb0f0s:
```suggestion
#if NET6_0_OR_GREATER
        if (username is ['@', ..])
#else
        if (username.Length > 1 && username[0]=='@')
#endif
```

tuscen (approved):

Issue #1112, opened 2022-06-24
Middleware to validate secret_token


karb0f0s:
We can ship middleware/extension method to validate [`secret_token`](https://core.telegram.org/bots/api#setwebhook) in ASP.NET Core applications.

> If you'd like to make sure that the webhook was set by you, you can specify secret data in the parameter secret_token. If specified, the request will contain a header “X-Telegram-Bot-Api-Secret-Token” with the secret token as content.

wiz0u:
Some people start reproaching ASP.NET dependencies on our main library
https://github.com/TelegramBots/book/issues/112

So I put this issue on hold as we might more all ASP.NET related code in a separate package in the future

wiz0u:
Why make things complex when it can be solved by adding a single line in the controller:
```csharp
    [HttpPost]
    public async Task<IActionResult> Post([FromBody] Update update)
    {
        if (Request.Headers["X-Telegram-Bot-Api-Secret-Token"] != SecretToken) return Forbid();
        // handle the update
    }
```

see https://github.com/TelegramBots/Telegram.Bot.Examples/blob/master/Webhook.Controllers/Controllers/BotController.cs#L24

PR #1190, opened 2023-01-11
Created Secret token validator middleware

BlueXTX:
resolve issue #1112

### DIFF
--- a/Telegram.Bot.sln
+++ b/Telegram.Bot.sln
@@ -47,6 +47,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".devcontainer", ".devcontai
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "EnumSerializer.Generator", "src\EnumSerializer.Generator\EnumSerializer.Generator.csproj", "{C246B54F-018B-4589-B26D-C38D7846FFF2}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Telegram.Bot.AspNet", "src\Telegram.Bot.AspNet\Telegram.Bot.AspNet.csproj", "{93755AE6-F438-4D86-9C36-A54A32549F6D}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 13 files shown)


BlueXTX:
resolve #1190

karb0f0s:
First of all, @BlueXTX thank you for your contribution!
Though I have a few concerns about this change:
* The biggest issue is ASP.NET dependencies. I'm not sure how they play out for legacy platforms, covered by .netstandard2.0 target. It might be safer to introduce this change under `#if DOTNET6_OR_GREATER` directive to avoid collision with older .NET frameworks.
* I don't see a way to differentiate between controllers for this middleware. If I have other endpoints in my app, they would require `X-Telegram-Bot-Api-Secret-Token` header too.
* IMO helper functions and middleware classes would greatly benefit from some level of XML documentation on them.

@tuscen WDYT?

BlueXTX:
> First of all, @BlueXTX thank you for your contribution! Though I have a few concerns about this change:
> 
> * The biggest issue is ASP.NET dependencies. I'm not sure how they play out for legacy platforms, covered by .netstandard2.0 target. It might be safer to introduce this change under `#if DOTNET6_OR_GREATER` directive to avoid collision with older .NET frameworks.
> * I don't see a way to differentiate between controllers for this middleware. If I have other endpoints in my app, they would require `X-Telegram-Bot-Api-Secret-Token` header too.
> * IMO helper functions and middleware classes would greatly benefit from some level of XML documentation on them.
> 
> @tuscen WDYT?

1. [Microsoft.Extensions.Options](https://www.nuget.org/packages/Microsoft.Extensions.Options/#supportedframeworks-body-tab) and [Microsoft.AspNetCore.Http.Abstractions](https://www.nuget.org/packages/Microsoft.AspNetCore.Http.Abstractions/#supportedframeworks-body-tab) have support for netstandard2.0. 
2. Will it be enough if I add configuration like this? `app.UseWhen(context => context.Request.Path.StartsWithSegments("/YourPath"), appBuilder =>
        {
            appBuilder.UseMiddleware<SecretTokenValidatorMiddleware>();
        });`

tuscen:
I don't think we should take dependency on M.E.x libraries in the core package. It's ok to make a separate package for ASP.NET Core integration though.

wiz0u:
Thanks for your contribution @BlueXTX 
For now, I think it's simpler to do without such middleware (see #1112)

PR #1298, opened 2023-08-19
Message id type changed to int

Elizerph:
Message id type changed from string to int for sake of consistency of Message class id property


karb0f0s:
you cannot randomly assign types to properties. they all follow (more or less) schema, defined in official [Telegram Bot API](https://core.telegram.org/bots/api)

Issue #1334, opened 2024-01-03
Bot API 7.0 (December 29, 2023)


karb0f0s:
# [Bot API 7.0 (December 29, 2023)](https://core.telegram.org/bots/api#december-29-2023)

## Reactions

Added the classes [ReactionTypeEmoji](https://core.telegram.org/bots/api#reactiontypeemoji) and [ReactionTypeCustomEmoji](https://core.telegram.org/bots/api#reactiontypecustomemoji) representing different types of reaction.
Added updates about a reaction change on a message with non-anonymous reactions, represented by the class [MessageReactionUpdated](https://core.telegram.org/bots/api#messagereactionupdated) and the field message_reaction in the class [Update](https://core.telegram.org/bots/api#update). The bot must explicitly allow the update to receive it.
Added updates about reaction changes on a message with anonymous reactions, represented by the class [MessageReactionCountUpdated](https://core.telegram.org/bots/api#messagereactioncountupdated) and the field message_reaction_count in the class [Update](https://core.telegram.org/bots/api#update). The bot must explicitly allow the update to receive it.
Added the method [setMessageReaction](https://core.telegram.org/bots/api#setmessagereaction) that allows bots to react to messages.
Added the field available_reactions to the class [Chat](https://core.telegram.org/bots/api#chat).

## Replies 2.0

Added the ability to reply to messages in other chats or forum topics.
Added the class [ExternalReplyInfo](https://core.telegram.org/bots/api#externalreplyinfo) and the field external_reply of type [ExternalReplyInfo](https://core.telegram.org/bots/api#externalreplyinfo) to the class [Message](https://core.telegram.org/bots/api#message), containing information about a message that is replied to by the current message, but can be from another chat or forum topic.
Added the ability to quote a part of the replied message.
Added the class [TextQuote](https://core.telegram.org/bots/api#textquote) and the field quote of type [TextQuote](https://core.telegram.org/bots/api#textquote) to the class Message, which contains the part of the replied message text or caption that is quoted in the current message.
[...]

PR #1335, opened 2024-01-03
Bot API 7.0 and 7.1

karb0f0s:
Implements #1334


src/Telegram.Bot/Types/ReactionType.cs
```diff
@@ -0,0 +1,62 @@
...
+[JsonObject(MemberSerialization.OptIn, NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
+public class ReactionTypeEmoji : ReactionType
+{
+    /// <summary>
+    /// Type of the reaction, always "emoji"
+    /// </summary>
+    public override string Type => "emoji";
+
+    /// <summary>
+    /// Reaction emoji. Currently, it can be one of "👍", "👎", "❤", "🔥", "🥰", "👏", "😁",
```
tuscen, line 35: I think we should add a static class with const fields for each emoji to make it more discoverable. You can look names for each emoji here http://emojipedia.org
```csharp
public static class KnownReactionTypeEmoji
{
    public const string ThumbsUp = "👍";
    // etc
}
```

src/Telegram.Bot/Types/Update.cs
```diff
@@ -119,20 +136,22 @@ public class Update
...
+        { EditedChannelPost: { } }    => UpdateType.EditedChannelPost,
+        { ShippingQuery: { } }        => UpdateType.ShippingQuery,
+        { PreCheckoutQuery: { } }     => UpdateType.PreCheckoutQuery,
+        { Poll: { } }                 => UpdateType.Poll,
+        { PollAnswer: { } }           => UpdateType.PollAnswer,
+        { MyChatMember: { } }         => UpdateType.MyChatMember,
+        { ChatMember: { } }           => UpdateType.ChatMember,
+        { ChatJoinRequest: { } }      => UpdateType.ChatJoinRequest,
+        { MessageReaction: { } }      => UpdateType.MessageReaction,
+        { MessageReactionCount: { } } => UpdateType.MessageReactionCount,
```
tuscen, line 154: Can you replace `{ }` to `not null` in each arm like so `{ Message: not null }`? It reads more explicitly like that.

src/Telegram.Bot/Types/ReactionType.cs
karb0f0s: I'll die on that hill :(

src/Telegram.Bot/Requests/Available methods/Messages/SetMessageReactionRequest.cs
```diff
@@ -0,0 +1,57 @@
...
+    [JsonProperty(Required = Required.Always)]
+    public int MessageId { get; }
+
+    /// <summary>
+    /// New list of reaction types to set on the message. Currently, as non-premium users, bots can
+    /// set up to one reaction per message. A custom emoji reaction can be used if it is either
+    /// already present on the message or explicitly allowed by chat administrators.
+    /// </summary>
+    [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+    public ReactionType[]? Reaction { get; set; }
```
tuscen, line 33: Change it to `IEnumerable<ReactionType>`, please. We use it everywhere in the inputs

src/Telegram.Bot/TelegramBotClientExtensions.ApiMethods.cs
```diff
@@ -1804,6 +1804,49 @@ await botClient.ThrowIfNull()
...
+    /// Pass <see langword="true"/> to set the reaction with a big animation
+    /// </param>
+    /// <param name="cancellationToken">
+    /// A cancellation token that can be used by other objects or threads to receive notice of cancellation
+    /// </param>
+    public static async Task SetMessageReactionAsync(
+        this ITelegramBotClient botClient,
+        ChatId chatId,
+        int messageId,
+        ReactionType[]? reaction,
```
tuscen, line 1836: Here's also `IEnumerable<ReactionType>`

src/Telegram.Bot/Types/ReactionType.cs
```diff
@@ -0,0 +1,62 @@
+using Telegram.Bot.Converters;
+
+namespace Telegram.Bot.Types;
+
+/// <summary>
+/// This object describes the type of a reaction. Currently, it can be one of
+/// <list type="bullet">
+/// <item><description>ReactionTypeEmoji</description></item>
+/// <item><description>ReactionTypeCustomEmoji</description></item>
```
tuscen, line 9:
```suggestion
/// <item><description><see cref="ReactionTypeEmoji"/></description></item>
/// <item><description><see cref="ReactionTypeCustomEmoji"/></description></item>
```

src/Telegram.Bot/Types/ReactionType.cs
karb0f0s: addressed

src/Telegram.Bot/Extensions/HttpContentExtensions.cs
```diff
@@ -21,6 +21,7 @@ internal static class HttpContentExtensions
 
         // It will be dispose of after the request is made
 #pragma warning disable CA2000
+        inputFile.Content.Position = 0;
```
tuscen, line 24: Not every stream is rewindable. We need to think of another solution

tuscen: One idea is to make a buffer stream using `MemoryStream`, though I'm not sure how to deal with disposing of that stream

karb0f0s: What if we cache file content in Lazy<whatever> field and reuse it on consequent calls?

karb0f0s: So, I was able to cache content, but it'll require quite a significant change:

```csharp
public ctor()
{
    cachedStreamContent = new(Do);
}

private readonly Lazy<Task<StreamContent?>> cachedStreamContent;

private async Task<StreamContent?> Do()
{
    HttpContent? content =  ...;
    var stream = await content!.ReadAsStreamAsync().ConfigureAwait(false);
    return new StreamContent(stream);
}

public override Task<StreamContent?> HttpContent() => cachedStreamContent.Value;
```

BTW .NET 6+ has synchronous `ReadAsStream()` method.

karb0f0s: #1339

karb0f0s: #1341

smetlov:
Got the ETA for this one?

tuscen (dismissed):

CumpsD:
Looking forward to be able to use Reactions! :)

tuscen (approved):

CumpsD:
Any idea when this will be released?

tuscen:
@CumpsD https://github.com/TelegramBots/Telegram.Bot/issues/1375

CumpsD:
> @CumpsD #1375

Thanks! Solved it by downloading the nupkg from releases and adding this to my csproj:

```
    <PropertyGroup>
        <RestoreSources>$(RestoreSources);../assets;https://api.nuget.org/v3/index.json</RestoreSources>
    </PropertyGroup>
```

(nupkg is stored in folder `assets`)

PR #1368, opened 2024-03-27
Replace Newtonsoft.Json with System.Text.Json

tuscen:


Fedorus:
ApiResponse<TResult> contains private setters that does not allow deserialization of any object

PR #1371, opened 2024-03-31
Implement Bot API 7.2

tuscen:


src/Telegram.Bot/Requests/Stickers/SetStickerSetThumbnailRequest.cs
```diff
@@ -49,6 +56,7 @@ public SetStickerSetThumbnailRequest(string name, long userId)
     {
         Name = name;
         UserId = userId;
+        Format = StickerFormat.Static;
```
BloodShadow174, line 59: hi
shouldn't the format be put into the constructor?

PR #1377, opened 2024-05-07
AOT support

notelgnis:
## Description
This pull request introduces support for Ahead-Of-Time (AOT) compilation in the library. As the original branch has already had `System.Text.Json` serialisation implemented, the main effort was to make all the API requests and types involved in communication AOT friendly and to remove everything that wasn't AOT friendly. This introduces another source generator in addition to `EnumSerializer.Generator` which is standard `System.Text.Json` source generator. 

The library API has been preserved, however, if you for any reason want to serialise or deserialise a request manually in your code (say you get a raw input from your web hook) you'll have to use `TelegramBotClientJsonSerializerContext` like this:

```csharp
Update update =  
    JsonSerializer.Deserialize(rawTelegramUpdate.Update, TelegramBotClientJsonSerializerContext.Instance.Update);
```

 The `TelegramBotClientJsonSerializerContext` contains all the generated code needed for serialisation and deserialisation. Another thing worth mentioning is this `TelegramBotClientJsonSerializerContext.Instance` static member which should be used instead of default implementation. The standard implementation of `JsonSerializerContext` generator generates `TelegramBotClientJsonSerializerContext.Default` but it lacks some important `JsonSerializerOptions` like `JsonNamingPolicy.SnakeCaseLower` etc. The `Instance` instance has them. You can also change  default `JsonSerializerOptions`  in `Instance` like this:
```csharp
TelegramBotClientJsonSerializerContext.JsonSerializerOptions = new JsonSerializerOptions { // your options }
```

However, if you just use the library API methods for sending or polling requests you won't need this and there are no changes needed in your code as well. 
 
## Changes
There are some major internal changes in the library though. First comes from the necessity of having two source generators and the fact that the one depends on the output of the other. Two of them don't properly work in the same project so the library has been split in two: `Telegram.Bot` and `Telegram.Bot.Base`. The last one includes the code generated by `EnumSerializer.Generator`and the first the code generated by `System.Text.Json` source generator. So, there will be most likely two Nuget packages as well.

Other modifications include removing `JsonSerializerOptionsProvider` and some convenient attributes like  `CustomJsonDerivedTypeAttribute`, `CustomJsonPolymorphicAttribute`,  `PolymorphicJsonConverter`,`PolymorphicJsonConverterFactory` as they weren't AOT friendly.

### Unit Tests Changes
Unit tests had to be changed as well. To test the same serialisation context that is used in actual API calls it was necessary to test serialisation with the actual (not container-like) types.


codeinhabitant:
Another thing I forgot to mention is that AOT warnings only start appearing in .NET 7. However, the target frameworks have been left unchanged for compatibility reasons. Therefore, to check for any new AOT compatibility warnings, it is necessary to temporarily switch the target framework to .NET 7 or higher.

wiz0u:
Honestly, I'm not sure I will use this PR. It's based on old branch, old approach to STJ. Your PR is hard to read because of lots of change in all files plus an extra Telegram.Bot.Base folder that duplicate all files (not sure of the reason to keep those)

Anyway, I'm pretty sure we can achieve the same without adding TelegramBotClientJsonSerializerContext.Instance.* arg to every constructor. My experience of doing it in WTelegramClient seemed much simpler.
Also I can probably simplify things with my code generation and maybe source generator.

So I will probably investigate this issue myself and find a good simpler way to achieve AOT/Trimming support from current codebase.

notelgnis:
Yeah, this PR is pretty old and I probably would have done it differently myself now.

PR #1394, opened 2024-07-05
Minor changes to help with WTelegramBot compatibility

wiz0u:
The only breaking change is BotId becoming a long, but I don't feel like making a 21.5 just for that.


src/Telegram.Bot/Polling/DefaultUpdateReceiver.cs
```diff
@@ -8,41 +8,32 @@ namespace Telegram.Bot.Polling;
...
-    /// <param name="botClient">The <see cref="ITelegramBotClient"/> used for making GetUpdates calls</param>
-    /// <param name="receiverOptions">Options used to configure getUpdates requests</param>
-    public DefaultUpdateReceiver(
-        ITelegramBotClient botClient,
-        ReceiverOptions? receiverOptions = default)
-    {
-        _botClient = botClient ?? throw new ArgumentNullException(nameof(botClient));
-        _receiverOptions = receiverOptions;
-    }
+    readonly ITelegramBotClient _botClient = botClient ?? throw new ArgumentNullException(nameof(botClient));
```
tuscen, line 24: You use `ReceiverOptions` without a separate private readonly field. Why not do the same with the bot client?

tuscen (approved):

PR #1427, opened 2024-11-01
Bot API 7.11 + Serialization code revamp + Methods renamed

wiz0u:
Bot API 7.11 + Serialization code revamp + Methods renamed


src/Telegram.Bot/ITelegramBotClient.cs
```diff
@@ -58,7 +58,7 @@ public interface ITelegramBotClient
     /// <param name="request">API request object</param>
     /// <param name="cancellationToken"></param>
     /// <returns>Result of the API request</returns>
-    Task<TResponse> MakeRequestAsync<TResponse>(
+    Task<TResponse> MakeRequest<TResponse>(
```
tuscen, line 61: What do you think about renaming it to `SendRequest`? I've long thought about it since `MakeRequest` doesn't sound right to me.

src/Telegram.Bot/Requests/FileRequestBase.cs
```diff
@@ -1,71 +1,45 @@
...
-    protected MultipartFormDataContent ToMultipartFormDataContent(
-        string fileParameterName,
-        InputFileStream inputFile)
+    /// <inheritdoc/>
+    public override HttpContent? ToHttpContent()
     {
-        if (inputFile is null or { Content: null })
+        InputFileConverter.Attachments.Value = null;
+        var utf8Json = JsonSerializer.SerializeToUtf8Bytes(this, GetType(), JsonBotAPI.Options);
+        var attachments = InputFileConverter.Attachments.Value;
```
tuscen, line 17: I'm not sure if this is a good idea to make such a tight coupling between `InputFileConverter` and `FileRequestBase`. It's really hard to follow such code.

src/Telegram.Bot/Requests/FileRequestBase.cs
```diff
@@ -1,71 +1,45 @@
...
+        if (attachments == null)
+            return new ByteArrayContent(utf8Json) { Headers = { ContentType = new("application/json") { CharSet = "utf-8" } } };
+
+        var multipartContent = new MultipartFormDataContent();
+        var firstLevel = JsonDocument.Parse(utf8Json).RootElement.EnumerateObject();
```
tuscen, line 22: Why do you need to first serialize and then to deserialize the JSON back and forth?

src/Telegram.Bot/ITelegramBotClient.cs
wiz0u: Interesting idea.
It's just that the v22 migration page helps remove the Async suffix via 3 search/replace regex, so it would make another 4th search/replace just for this... 🤔

src/Telegram.Bot/Requests/FileRequestBase.cs
wiz0u: It doesn't seem so illogical to have a coupling between those, they work toward the same goal. still makes things much simpler, avoid the many ToHttpMethod

src/Telegram.Bot/Requests/FileRequestBase.cs
wiz0u: That's actually what the previous STJ code we used did behind the scenes... I analyzed it because I was worried too.
In fact the new approach is even more efficient.

tuscen (approved):

Issue #1322, opened 2023-12-11
Support AOT/Trimming/Blazor WebAssembly using source generation


cyrmax:
## I would like to request/propose a feature

I suggest replacing NewtonSoft.Json with System.Text.Json with proper source generation used.
The reason is that with newtonsoft we cannot use Aot and/or trimming which then breaks the entire Json package.
With System.Text.Json we can see the similar issue with trimming and Aot. But there we can use source generation aproach and avoid any errors when trimming the executable or building native code.
Here is the documentation for this source generation feature in detail and System.Text.Json docs can be googled easily.
https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation?pivots=dotnet-8-0

If anybody has any arguments against my feature request I would be glad to discuss them.
Also I will try to implement the thing I want by myself and if I achieve what I want without any errors I of course will make a pull request to address this issue.

Thanks in advance.

karb0f0s:
@cyrmax wow, what a great idea! How would you approach it with current codebase?

tuscen:
Will this work for `netstandard2.0` targets?

cyrmax:
@tuscen 
I found the following information regarding your question.
TLDR: yes.
But here is a quote and a link.

> The System.Text.Json library is included in the runtime for .NET Core 3.1 and later versions. For other target frameworks, install > the System.Text.Json NuGet package. The package supports:
> • .NET Standard 2.0 and later versions
> • .NET Framework 4.7.2 and later versions
> • .NET Core 2.0, 2.1, and 2.2

And here is the link:
https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-8-0

tuscen:
@cyrmax Thanks, we'll probably try to migrate to source generated converters then

cyrmax:
@karb0f0s 
If you mean that it is impossible or too difficult then please explain why.
I didn't looked at the code yet but I think at least it is possible and reasonable.

tuscen:
@cyrmax By the way, you really don't want this library to be trimmed since it's an API client. What will it do in case your bot receives a type that was trimmed during compilation?

tuscen:
@cyrmax Do you have experience with STJ source generators? I’ve tried to use it in this library, but for some reason it doesn’t generate converters. But in a separate test project I don’t have any problems.

cyrmax:
[...]

PR #1435, opened 2024-11-13
Actual SDK support

MoryakSPb:
As of November 12, .NET 6 is no longer supported by Microsoft. It is necessary to provide support for currently relevant SDKs.

### DIFF
--- a/.azure-pipelines/variables.yml
+++ b/.azure-pipelines/variables.yml
@@ -1,7 +1,7 @@
 variables:
   - group: Integration Tests Variables
   - name: versionPrefix
-    value: 22.0.2
+    value: 22.0.3
   - name: versionSuffix
     value: ''
   - name: ciVersionSuffix
@@ -19,7 +19,7 @@ variables:
     ${{ else }}:
       value: $(versionPrefix)-$(ciVersionSuffix)
   - name: netSdkVersion
-    value: 8.0.x
+    value: 9.0.x
   - name: buildConfiguration
     value: Release
   - name: vmImage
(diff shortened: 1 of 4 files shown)


wiz0u:
Our .NET 6 build is compatible with projects using any more recent .NET projects. No need to produce specific build variants of our library unless we want to take advantage of some features in this variant only.

See the color legend:
image

MoryakSPb:
Yes, right now it is not a problem and we can use the library in compatibility mode. Problems will arise in the future when dependencies stop supporting .NET 6. The update will also reduce the number of transitive dependencies: `System.Text.Json` in version 9.0.0 has no dependencies. This version exludes support for .NET 6, so when building .NET 6, `System.Text.Json` for .NET Standard 2.0 will be used.

MoryakSPb:
@tuscen already had a problem with targeting .NET 6:
https://github.com/TelegramBots/Telegram.Bot/issues/1322#issuecomment-1861240327

wiz0u:
I will add two things:
- System.Text.Json 9.0 is already included in .NET 9 base framework, so no need to include it explicitly for such target, projects for .NET 9 already use that version.
- If you want to take advantage of System.Text.Json 9.0 in your bot project for older targets, you can include it explicitly as dependency in your own project and Telegram.Bot will use it as well AFAIK

I will close this PR for now, as I feel adding specific targets is not really necessary.
Of course if the conditions are changing in the future (for example in order to properly support AOT), i will add adequate targets if it becomes necessary.

PR #1451, opened 2024-12-28
Expose bot token

7Bpencil:
Bot token is required for WebApp.initData validation and it would be convenient to get token directly from client instance instead of passing token along it into different services. Are there any issues with exposing bot token?


wiz0u:
Already done in next version.

PR #1483, opened 2026-06-16
perf: reduce allocations in PassportDecrypter

rashidiic:
Two hot paths in **PassportDecrypter** were allocating more than needed. Fixed both.

- DecryptDataStreamAsync: paddingBuffer now rents from ArrayPool<byte>.Shared instead of allocating on every call.
- FindDataKeyAndIv: swapped the temp hash buffer for IncrementalHash + stackalloc on .NET 5+. Added #if NETSTANDARD2_0 for the old path.

Benchmarks (BenchmarkDotNet, .NET 8, Apple M4 Pro):

| Method                                   |    Mean    | Allocated  |  Alloc Ratio  |
|-----------------------------|----------|-----------|-------------|
| Baseline_FindDataKeyAndIv | 211.7 ns   | 480 B       | 1.00              |
| Optimized_FindDataKeyAndIv  | 196.9 ns | 216 B     | 0.45             |

Allocations cut by ~55%. Gen0 collections per 1000 ops dropped from 0.0572 to 0.0257.

Benchmark project is in benchmarks/Telegram.Bot.Benchmarks — run with dotnet run -c Release.
Tests: 180 passed, 0 failed (.NET 8).


wiz0u:
This feels like an AI-generated PR just to get your name associated with this package.
Your benchmark shows a gain of just 15 ns for a method that is rarely called.
Your new code has limited interest and sometimes questionable choice.

PR #1490, opened 2026-09-20
Introduce UpdateHandlingException for update handling errors

Rikitav:
Wrap exceptions thrown during HandleUpdateAsync in DefaultUpdateReceiver into UpdateHandlingException, preserving the original exception as the inner exception and the Update that was being processed.

This allows error handlers to inspect the update that caused the failure without losing the original exception context.


wiz0u:
There are many ways to proceed otherwise:
- let the user add a big try..catch in their HandleUpdate to wrap the exception
- just save the latest update seen rather than wrap it (HandleError is called immediately after an exception)
- pass your own updateHandler that does this and call an inner updateHandler
- make your own IUpdateHandler instead of DefaultUpdateHandler

Why does this need to be in the library?
This would also break existing code from users expecting an unwrapped exception

Rikitav:
> There are many ways to proceed otherwise    

Indeed, but this is could be pretty much part of default handler behaviour.

> Why does this need to be in the library?    

This feels like a feature that should have been in library from very start.

> This would also break existing code from users expecting an unwrapped exception    

Existing bots usually aren't updated that frequently. As a last resort, this could be the saved feature for v23 as a breaking change.

wiz0u:
> This feels like a feature that should have been in library from very start.
HandleErrorAsync is used in 3 contexts (see HandleErrorSource), only one of these could have an update as added context.

Having the exception wrapped in only one of these context and not in the others seems weird for the bot developer standpoint. It's also a bit annoying from a memory standpoint, to create an exception instance just for that.

I find it simpler to save the last update in a field in updateHandler, so it can be used synchronously (on the same task) in the errorHandler on exception. But this is something that can be done easily by the bot dev (or by a framework)

I don't feel it's useful in the library right now. But I will reconsider it if we ever move to a v23 breaking version.

